PR #615, opened 2017-02-14
 Improved UX of Expression editor

roonyh:
* Reduced empty spaces around the text box
* Change colors
* Pressing enter sets the value

### DIFF
--- a/modules/web/js/expression-editor/expression-editor-utils.js
+++ b/modules/web/js/expression-editor/expression-editor-utils.js
@@ -15,21 +15,41 @@
  */
 define(['require', 'lodash', 'jquery'], function (require, _, $) {
     var expressionEditorUtil = {};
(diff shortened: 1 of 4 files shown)


modules/web/js/expression-editor/expression-editor-utils.js
```diff
@@ -15,26 +15,46 @@
...
+            return hiddenSpan.outerWidth();
+        }
+
+        $(propertyInputValue).css("width", getNecessaryWidth(propertyValue));
+
+        $(propertyInputValue).on("paste keyup", function (e) {
+            var width = getNecessaryWidth($(this).val());
+            $(this).css("width", width);
+
+            if(e.keyCode==13){
```
hemikak: Need spacing around the equals. And better to use ===.

roonyh: We are using an ace-editor instance for expression editor. This code seems to be outdated.

PR #653, opened 2017-02-15
Fix the issue where connector declaration title cannot be renamed when inside a connector definition.

roonyh:


modules/web/js/ballerina/views/connector-declaration-view.js
```diff
@@ -184,5 +184,24 @@ define(['lodash','d3', 'jquery', './ballerina-view', './../ast/connector-declara
             d3Element.attr("stroke-width", 0);
         };
 
+        ConnectorDeclarationView.prototype.createPropertyPane = function() {
```
hemikak: Shall we have a function comment here ?

roonyh: We have removed the `connector-declaration-view` module since we moved to React and replaced it with `connector-declaration.jsx` module. I will look into adding necessary doc comments to the instance methods there.

PR #908, opened 2017-02-18
Expression editor shows correct info for current package connector definitions

roonyh:
Fixes #677


modules/web/js/ballerina/ast/connector-declaration.js
```diff
@@ -250,6 +250,11 @@ define(['lodash', './node', '../utils/common-utils'], function(_, ASTNode, Commo
         return expression;
     };
 
+    var shouldSkipPackageName = function(packageName) {
+        return _.isUndefined(packageName) || _.isNil(packageName) ||
```
nadeeshaan: No need to check with both isUndefined and  isNil. isNil checks for both null and undefined

PR #1055, opened 2017-02-20
Fixing packages import getting distorted when importing more

### LINKED ISSUE #983: Import section get distorted when more than three imports are added
Please see the attached image, the import section is distorted when more than three imports added. 
composer_bug_imports_1

this:
Fixes #983


hemikak:
Fixes https://github.com/ballerinalang/composer/issues/1069 as well I presume ?

Issue #1057, opened 2017-02-21
Saving a file does not add .bal extension 


raviu:
Saving a file from the composer does not add the .bal extension. This is ok but it seems if you open a file without .bal extension you can't run/debug it on the composer. 

ENV: macOS 10.12.3 JDK 1.8.0 Google Chrome 56

dakshika:
Duplicate issue https://github.com/ballerinalang/composer/issues/132

nadeeshaan:
Closing since the fix #1060 is merged

PR #1060, opened 2017-02-22
Change version of composer and ballerina dependency version to 0.8.1-SNAPSHOT and fix the issue #1057

NipunaMarcus:
Added pom version changes and fix the issue: https://github.com/ballerinalang/composer/issues/1057


hemikak:
Have to do these changes as well : 
https://github.com/ballerinalang/composer/pull/934/files#diff-b19e6bcf97654e2339af2ad94fde96df
https://github.com/ballerinalang/composer/pull/934/files#diff-c3a502bd2bbd8ae98a172e7fce807970

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/app/WorkspaceServiceRunner.java
```diff
@@ -37,7 +37,7 @@
 /**
  * Workspace Service Entry point.
  *
- * @since 0.8.0
+ * @since 0.8.1
```
maheeka: This documentation is correct right? this was added for 0.8.0

NipunaMarcus: No this was mistakenly changed .... i will change it back to 0.8.0

PR #1065, opened 2017-02-22
add support to generate source from model

ruks:


hevayo:
Closing the PR since the implementation is differed.

Issue #1053, opened 2017-02-20
Argument/Parameter view is incorrect


maheeka:
<img width="521" alt="image" src="https://cloud.githubusercontent.com/assets/1029806/23130559/f30d1a58-f7ac-11e6-90fb-6c7a852b2cf0.png">
<img width="355" alt="image" src="https://cloud.githubusercontent.com/assets/1029806/23210788/1f7574c2-f925-11e6-98dc-33ee9cb7e8cf.png">

Named Return Types check should come first.

maheeka:
Before applying select2 : 
<img width="449" alt="image" src="https://cloud.githubusercontent.com/assets/1029806/23291375/1814af36-fa7e-11e6-8325-5dbdb18816d8.png">
<img width="331" alt="image" src="https://cloud.githubusercontent.com/assets/1029806/23291379/1ea7b26c-fa7e-11e6-9ea5-2407a3001588.png">

With the fix : 
<img width="494" alt="image" src="https://cloud.githubusercontent.com/assets/1029806/23294418/6214b564-fa91-11e6-891e-64240eebcd59.png">
<img width="379" alt="image" src="https://cloud.githubusercontent.com/assets/1029806/23294424/6a89c91e-fa91-11e6-9732-72b2969c2d69.png">

maheeka:
Fixed with https://github.com/ballerinalang/composer/pull/1082/commits/ceb19dffec6da814b78572f58e7fed7ce84e8517

maheeka:
Reproduced in latest master. Fix is broken by later commits.

maheeka:
Fixed with https://github.com/ballerinalang/composer/pull/1095/commits/af2ff446b73deead683d7511e4de8250b8de5125

PR #1082, opened 2017-02-24
fix return param panel element ordering issue : https://github.com/ba…

maheeka:
…llerinalang/composer/issues/1053


hemikak:
Fixes : https://github.com/ballerinalang/composer/issues/1053

PR #1116, opened 2017-03-01
Fixing minor positioning issues

nadeeshaan:
This fixes the following issue,

Re positioning of the worker receiver after deleting an intermediate statement
Unnecessarily setting the activated target on worker receiver hover


modules/web/js/ballerina/views/reply-statement-view.js
```diff
@@ -188,6 +188,11 @@ define(['lodash', 'log', './simple-statement-view', './../ast/reply-statement',
                 previousStatement = statementContainer.getManagedStatements()[currentIndex - 1];
                 previousStatementView = self.getDiagramRenderingContext().getViewOfModel(previousStatement);
                 return newBBoxTop >= previousStatementView.getBoundingBox().getBottom() + innerDropZoneHeight;
+            } else if (currentIndex === 0) {
+                return true;
+            } else {
+                log.error("Invalid index Found for the Reply Statement");
```
NipunaMarcus: why do we throw and log? can't we do log or throw?

nadeeshaan: Basically, log.error is used since it will log the error in our log file while throw is used since I need to stop the whole execution since index being -1 due to a serious positioning issue

Issue #1071, opened 2017-02-23
Graphical veiw doesn't allow to rename a function /service - on firefox


kavithasub:
Checked on 0.8 GA tools pack
 Firefox 51.0; Ubuntu 15.10

![funname](https://cloud.githubusercontent.com/assets/15671620/23248568/f85d9b98-f9c5-11e6-9ccb-478f9a2e362c.jpeg)

![sername](https://cloud.githubusercontent.com/assets/15671620/23248570/fa01afac-f9c5-11e6-90c8-f3d8ef17b1e6.jpeg)

NipunaMarcus:
This is happening because of Firefox ASCII system for recognizing characters is bit different as it recognize backspace as a special key. will fix this.

nadeeshaan:
Closing since the fix is merged

PR #1118, opened 2017-03-01
Fix the issue on Firefox where service/function/connector names doesn't change

NipunaMarcus:
This fix will listen to the keypress() event and then check whether pressed keys are Delete or Backspace. If it is not then proceed as before and if it is ignore these keys and let them do there default behaviors. This is only checked in Firefox as chrome and IE don't fire keypress() for Backspace or Delete.

Fix the issue: https://github.com/ballerinalang/composer/issues/1071

Note: Did not use the character code as Delete and "." has the same charcode


maheeka:
@NipunaMarcus , shouldn't this fix be extended to all other enter keypress events as well?

NipunaMarcus:
Enter keypress is fine..problem is in backspace and delete. I will check other sports were we use inline editing

NipunaMarcus:
Do not merge. There are changes to be added

NipunaMarcus:
Please review and merge

PR #1124, opened 2017-03-01
Add web resources as an artifact and remove distribution module

akalankapagoda:


akalankapagoda:
Closing this so that we can break adding web module and moving resources into two pull requests to avoid conflicts.

PR #1135, opened 2017-03-02
jsPlumb is not exported as default from 'jsplumb'

roonyh:
Changed the way its used.


modules/web/js/ballerina/views/ballerina-file-editor.js
```diff
@@ -808,13 +808,11 @@ define(['lodash', 'jquery', 'log', './ballerina-view', './service-definition-vie
...
                 return function findMatches(q, cb) {
-                    var matches, substringRegex;
-
                     // an array that will be populated with substring matches
-                    matches = [];
+                    var matches = [];
 
                     // regex used to determine if a string contains the substring `q`
-                    substringRegex = new RegExp(q, 'i');
+                    var substrRegex = new RegExp(q, 'i');
```
nadeeshaan: What if we use "let" instead of "var" since we use ES6 now

PR #1136, opened 2017-03-02
Fixing worker delete issue

nadeeshaan:
When we delete a worker when there are added connectors, this won't re position the connectors accordingly


nadeeshaan:
Resolves #1137

PR #1162, opened 2017-03-05
Add search box to ace editor

roonyh:
'ace-builds' defines some features of the ace editor in separate modules. So need to require them explicitly if we want them.

### DIFF
--- a/modules/web/js/ballerina/utils/ace-styles.css
+++ b/modules/web/js/ballerina/utils/ace-styles.css
@@ -1,0 +1,3 @@
+.tab-pane .ace_search {
+  font: 11px system-ui;
+}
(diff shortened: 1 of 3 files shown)


modules/web/js/ballerina/utils/ace-styles.css
```diff
@@ -0,0 +1,3 @@
+.ace_search {
```
kaviththiranga: Can we further reduce the scope of this selector? Say like inside a tab-pane

roonyh: Sure. Updated selector to `.tab-pane .ace-search`.

PR #1163, opened 2017-03-05
Executing npm commands using maven-exec-plugin.

hemikak:


kaviththiranga:
As discussed offline with @hemikak, for the moment we are making our primary builder maven - and will get necessary npm scripts executed via maven. However, in future, after electron related stuff comes into build, we may alter this and go with node as primary builder.

PR #1181, opened 2017-03-09
[Fixed #1180] - Build fails on windows

### LINKED ISSUE #1180: Build fails on windows
Build fails on window when running "npm" command on shell

jeradrutnam:


modules/web/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "clean": "rimraf dist",
     "build": "npm run webpack",
-    "webpack": "NODE_ENV=production webpack --config webpack.config.js",
+    "webpack": "cross-env NODE_ENV=production webpack --config webpack.config.js",
     "dev": "webpack-dev-server --hot --inline --progress",
     "runTests": "NODE_ENV=test mocha-webpack --require ./js/tests/js/spec/setup.js --webpack-config ./webpack.config.js ./js/tests/js/spec/BallerinaTest.js;",
```
hemikak: Can we set cross-env here as well ?

modules/web/pom.xml
```diff
@@ -136,6 +136,21 @@
...
+                <os>
+                    <family>windows</family>
+                </os>
+            </activation>
+            <properties>
+                <!-- Override the executable names for Windows -->
+                <npm.executable>npm.cmd</npm.executable>
+            </properties>
+        </profile>
+    </profiles>
```
hemikak: I think we need to add the default property as well : 

`	
I used this workaround to have a cross-platform Maven build : declare the npm executable name as a Maven variable, and use Maven filters to modify this executable name when running on Windows.

It can work the same for Grunt, Bower etc.

<properties>
    <npm.executable>npm</npm.executable>
</properties>`

http://stackoverflow.com/questions/22708255/cannot-run-program-npm-in-directory

Issue #85, opened 2017-02-09
[Editor] Annotation Pane and Argument pane for Resources does not position when multiple connectors are added.


uthaiyashankar:
<a href="https://github.com/hemikak"><img src="https://avatars.githubusercontent.com/u/4398820?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [hemikak](https://github.com/hemikak)**
_Saturday Jan 14, 2017 at 07:53 GMT_
_Originally opened as https://github.com/ballerinalang/ballerina/issues/772_

----

![image](https://cloud.githubusercontent.com/assets/4398820/21953428/94974696-da5c-11e6-98e8-94fd732c7ac7.png)

sumuditha-viraj:
PR #1177

sumuditha-viraj:
Fixed with #1177

sumuditha-viraj:
Fixes regarding connector action editor panes is with PR #1189

PR #1189, opened 2017-03-13
Fixing #85 - Part 2 - Fixing Annotation Pane and Argument pane issues…

sumuditha-viraj:
… for Connector action

### DIFF
--- a/modules/web/js/ballerina/views/function-arguments-view.js
+++ b/modules/web/js/ballerina/views/function-arguments-view.js
@@ -32,11 +32,14 @@ define(['require', 'lodash', 'jquery', 'ballerina/ast/ballerina-ast-factory'],
             var model = _.get(args, "model");
             var paneElement = _.get(args, "paneAppendElement");
             var viewOptions = _.get(args, "viewOptions");
+            this._viewOfModel = _.get(args, "view");
             var disableEditing = _.get(args, "disableEditing");
 
             var argumentsEditorWrapper = $("<div/>", {
                 class: "main-action-wrapper arguments-main-action-wrapper"
             }).appendTo(paneElement);
+
+            this._argumentsEditorWrapper = argumentsEditorWrapper;
 
             // Positioning the main wrapper
             argumentsEditorWrapper.css("left",
@@ -170,6 +173,8 @@ define(['require', 'lodash', 'jquery', 'ballerina/ast/ballerina-ast-factory'],
                     $(event.data.activatorElement).click();
                 }
             });
+
+            return this;
         };
 
         /**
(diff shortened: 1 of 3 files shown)


modules/web/js/ballerina/views/return-types-pane-view.js
```diff
@@ -293,5 +293,27 @@ define(['lodash', 'log', 'jquery', 'alerts', './return-type-view', './../ast/nod
             return dropdownData;
         };
 
+        /**
+         * moves the position of return type editor view
+         * @param {Object} args - object which contains delta values for x and y
+         * @param {Object} args.dx - delta value for x value
```
hemikak: if dx and dy are value, it should be {number} instead of {Object}

modules/web/js/ballerina/views/function-arguments-view.js
```diff
@@ -295,9 +300,32 @@ define(['require', 'lodash', 'jquery', 'ballerina/ast/ballerina-ast-factory'],
             return dropdownData;
         }
 
+        /**
+         * moves the position of argument editor view
+         * @param {Object} args - object which contains delta values for x and y
+         * @param {Object} args.dx - delta value for x value
+         * @param {Object} args.dy - delta value for y value
```
hemikak: if dx and dy are value, it should be {number} instead of {Object}

modules/web/js/ballerina/views/return-types-pane-view.js
```diff
@@ -293,5 +293,27 @@ define(['lodash', 'log', 'jquery', 'alerts', './return-type-view', './../ast/nod
...
+        ReturnTypePaneView.prototype.move = function (args) {
+            var dx = _.get(args, "dx", 0);
+            var dy = _.get(args, "dy", 0);
+
+            // Left margin of the connector action view
+            var leftMargin = this._viewOfModel.getBoundingBox().x();
+            var returnTypeEditorLeft = parseInt(this._returnTypeEditorWrapper.css("left"), 10) + dx;
+            // This is to ensure that return type editor view doesn't go beyond the left margin of the connector action.(to avoid clipping)
+            if (leftMargin < returnTypeEditorLeft) {
+                this._returnTypeEditorWrapper.css("left", returnTypeEditorLeft);
```
hemikak: I think we can use :
`this._returnTypeEditorWrapper.left(returnTypeEditorLeft);`

modules/web/js/ballerina/views/return-types-pane-view.js
sumuditha-viraj: Fixed.

modules/web/js/ballerina/views/function-arguments-view.js
sumuditha-viraj: Fixed.

PR #1221, opened 2017-03-17
When a statement is added select the relevant inner drop zone if not selected

### LINKED ISSUE #163: Delete and undo a statement will always add the statement to the end of timeline
<a href="https://github.com/maheeka"></img></a> **Issue by [maheeka](https://github.com/maheeka)**
_Thursday Jan 26, 2017 at 05:25 GMT_
_Originally opened as https://github.com/ballerinalang/ballerina/issues/1049_

----

When deleting a statement in a lifeline, the undo will always add the statement to as the last statement in timeline instead of its original place when deleted.

roonyh:
fixes #163

### DIFF
--- a/modules/web/js/ballerina/views/statement-container.js
+++ b/modules/web/js/ballerina/views/statement-container.js
@@ -92,11 +92,11 @@ define(['lodash', 'jquery', 'd3', 'log', 'd3utils', './point', './ballerina-view
 
     /**
      * Finds the index in the lifeLine where a statement should be rendered based on managed statements
-     * and the parents current children and selects the relevant inner drop zone.
+     * and the parent's current children and selects the relevant inner drop zone if possible.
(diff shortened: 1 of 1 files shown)


modules/web/js/ballerina/views/statement-container.js
```diff
@@ -89,6 +89,48 @@ define(['lodash', 'jquery', 'd3', 'log', 'd3utils', './point', './ballerina-view
...
+                    break;
+                }
+            }
+        }
+    }
+
+    /**
+     * Returns true if an inner drop zone is pending. That is if the user is hovering over statement with
+     * possible statement
+     * @returns {Boolean}
```
hemikak: @returns {boolean} true if an inner drop zone is pending, false otherwise.

roonyh: This code is outdated since we moved to React. This logic is now handled by React's re-rendering.

Issue #553, opened 2017-02-13
Icons in resources are not displayed in Firefox


dilinisg:
Pack - M4

![icons](https://cloud.githubusercontent.com/assets/1845370/22883153/c5cfa6ca-f214-11e6-94a4-20c41eb50191.png)

pahans:
this working in the latest build. firefox 52, 51.0 and chrome.

pahans:
not working on firefox 50

Issue #1017, opened 2017-02-20
Debug pointers are not shown in design view in Firefox


erandidesilva:
Firefox version: 48.0
Please see the attached image, the debug pointers are not shown in firefox.
![composer_bug_firefox_1](https://cloud.githubusercontent.com/assets/15624590/23111831/482aaf7a-f751-11e6-84ff-9b3b9fc417bc.png)

dakshika:
Seems working fine on latest FF 51.0.1

pahans:
tested this on firefox 51 with the latest build. It's working now. fixed delete button alignment issue in above screenshot with https://github.com/ballerinalang/composer/pull/1225

pahans:
fixed with https://github.com/ballerinalang/composer/pull/1236

hevayo:
Tested in 48 and fix is working.

Issue #300, opened 2017-02-09
[Editor] Delete icon of each statements in graphical view is not visible


uthaiyashankar:
<a href="https://github.com/ushanib"><img src="https://avatars.githubusercontent.com/u/12966170?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [ushanib](https://github.com/ushanib)**
_Monday Feb 06, 2017 at 11:34 GMT_
_Originally opened as https://github.com/ballerinalang/ballerina/issues/1465_

----

Delete icon of each statements in graphical view is not visible.

Please see the attachment.

![screenshot from 2017-02-06 17-02-10](https://cloud.githubusercontent.com/assets/12966170/22645701/52ffbd50-ec8e-11e6-8cc3-b81c8cbc8c59.png)

nadeeshaan:
Fixing since the fix is merged

Issue #133, opened 2017-02-09
[Editor] Expression edit selection PopUp load with empty content on Firefox browser


uthaiyashankar:
<a href="https://github.com/gvasanka"><img src="https://avatars.githubusercontent.com/u/3368635?v=3" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [gvasanka](https://github.com/gvasanka)**
_Monday Jan 23, 2017 at 13:03 GMT_
_Originally opened as https://github.com/ballerinalang/ballerina/issues/964_

----

Steps:
1. Drag and drop main function 
2. Click on system:println
3. check the loading popup
Issue:
On Firefox 50.1.0 Pop-Up loading with empty content. Check screenshots taken on Firefox and Chrome

![selection_135](https://cloud.githubusercontent.com/assets/3368635/22204859/3387e906-e19a-11e6-8c49-d47da531b82c.png)
![selection_136](https://cloud.githubusercontent.com/assets/3368635/22204861/3499f276-e19a-11e6-9506-a56828ea0017.png)

pahans:
related to https://github.com/ballerinalang/composer/pull/1225

pahans:
works on firefox 51 and 52 but not on 50.0.1

nadeeshaan:
closing since the fix is merged

PR #1236, opened 2017-03-21
fix svg icons not working on firefox 50.1

pahans:
fixes https://github.com/ballerinalang/composer/issues/553 , https://github.com/ballerinalang/composer/issues/300 and https://github.com/ballerinalang/composer/issues/133 

more info
http://stackoverflow.com/questions/15123953/svg-fill-url-behaving-strangely-in-firefox


pahans:
also fixes https://github.com/ballerinalang/composer/issues/1017

PR #1253, opened 2017-03-22
fix imports

pahans:


modules/web/js/ballerina/tool-palette/drag-drop-manager.js
```diff
@@ -21,14 +21,14 @@ import Backbone from 'backbone';
...
     {
-        idAttribute: this.cid,
         modelName: "DragDropManager",
         /**
          * @augments Backbone.Model
          * @constructs
          * @class Handles validations for drag and drop
          */
         initialize: function (attrs, options) {
+          this.idAttribute = this.cid;
```
hemikak: Is there an indentation issue here ?

pahans: this is fixed in latest master

PR #1256, opened 2017-03-23
Es6 changes

keizer619:


modules/web/js/type-mapper/type-mapper-renderer.js
```diff
@@ -22,16 +22,17 @@ import jsPlumbLib from 'jsplumb';
...
-     * Renderer constructor for TypeMapper
-     * @param {object} onConnectionCallback call back function when connection made
-     * @param {object} onDisconnectCallback call back function when connection removed
-     * @param {object} typeConverterView Type Mapper View reference object
-     * @constructor
-     */
-    var jsPlumb = jsPlumbLib.jsPlumb;
+/**
+ * Renderer constructor for TypeMapper
+ * @param {object} onConnectionCallback call back function when connection made
```
hemikak: If its a function ref, then it should be {function}

modules/web/js/type-mapper/type-mapper-renderer.js
```diff
@@ -22,16 +22,17 @@ import jsPlumbLib from 'jsplumb';
...
+ * @param {object} onDisconnectCallback call back function when connection removed
```
hemikak: If its a function ref, then it should be {function} instead of {object}

modules/web/js/type-mapper/type-mapper-renderer.js
```diff
@@ -986,18 +978,21 @@ import './type-mapper.css';
...
+    getConnectorConfig(midPoint) {
         return ["Flowchart", {
             midpoint: midPoint,
             stub: [40, 60], cornerRadius: 5, alwaysRespectStubs: true
-        }]
+        }];
     }
+}
+
+TypeMapperRenderer.prototype.constructor = TypeMapperRenderer;
```
hemikak: Do we need this ?

PR #1263, opened 2017-03-24
fix getChildContainer undefined error

pahans:


modules/web/js/ballerina/views/function-definition-view.js
```diff
@@ -212,9 +212,10 @@ class FunctionDefinitionView extends SVGCanvas {
         return this._viewOptions;
     }
 
-    getChildContainer() {
-        return this._childContainer ;
-    }
+    // TODO: this method this never used so is this._childContainer, remove them.
+    // getChildContainer() {
+    //     return this._childContainer ;
+    // }
```
hemikak: Shall we remove this comment ?

pahans: this file(`function-definition-view.js`) is removed from the repository due to migration of diagram views to react

PR #1264, opened 2017-03-24
Fixing the resource resizing on connector adding

nadeeshaan:
Fixes the issue #334


hemikak:
Might have to send a PR for es6-transition branch as well.

PR #1298, opened 2017-03-28
Fix code formatting issues.

roonyh:
* Fix adding a space between a tab and identifier
* Fix indentation not adding for nested empty block '}'s

### DIFF
--- a/modules/web/js/ballerina/utils/ace-formatter.js
+++ b/modules/web/js/ballerina/utils/ace-formatter.js
@@ -214,7 +214,7 @@ define(function (require) {
 
                 if(_.isEqual(token.type, 'ballerina-identifier')
                     && _.isEqual(lastToken.type, 'whitespace')
-                    && !_.endsWith(code, space)){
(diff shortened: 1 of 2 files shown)


modules/web/js/ballerina/utils/ace-formatter.js
```diff
@@ -258,12 +258,14 @@ define(function (require) {
...
-                                    // if block has no content don't indent so the closing '{' is not unnecessarily indented
-                                    for (var i = 0; i < indentation; i++) {
-                                        value += tab;
-                                    }
+                                if(token.type === 'paren.lparen' && token.value === '{' && nextToken.type === 'paren.rparen' && nextToken.value === '}') {
+                                    // if block has no content don't indent so the closing '}' is not unnecessarily indented
+                                    indent -= 1;
+                                }
+                                for (var i = 0; i < indent; i++) {
+                                    value += tab;
```
sumuditha-viraj: If the value of "indent" cannot be a minus value, we can write this as shown below avoiding the for loop.

value += (tab* indent);

PR #1340, opened 2017-04-06
Load native data-types dynamically

sumuditha-viraj:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/rest/BallerinaProgramService.java
```diff
@@ -0,0 +1,96 @@
+package org.ballerinalang.composer.service.workspace.rest;
```
maheeka: missing copyright

sumuditha-viraj: Fixed.

PR #1365, opened 2017-04-17
Change composer version from 0.86-SNAPSHOT to 0.87-SNAPSHOT

NipunaMarcus:
Ballerina lang haven't change the version to 0.87-SNAPSHOT  yet so this PR only contains version changes related to the composer.


pom.xml
```diff
@@ -186,7 +186,7 @@
     </build>
 
     <properties>
-        <ballerina.version>0.86</ballerina.version>
+        <ballerina.version>0.86-SNAPSHOT</ballerina.version>
```
NipunaMarcus: Have to change the ballerina version after ballerina runtime change it.

PR #1386, opened 2017-04-19
Fix the issue when adding annotation definitions name get duplicated

NipunaMarcus:


hevayo:
Closing the PR since the implementation is differed

PR #1460, opened 2017-05-03
Added stringified parameters/arguments/return types of function-defs …

hemikak:
…and resource-defs.


hemikak:
Merged in https://github.com/ballerinalang/composer/pull/1469/files

PR #1486, opened 2017-05-07
Create/Update struct variables

pahans:


modules/web/js/ballerina/components/create-struct.jsx
```diff
@@ -27,6 +27,12 @@ class CreateStruct extends React.Component {
         }
     }
 
+    handleSubmit() {
```
hemikak: Can we have doc comments ?

pahans: fixed with https://github.com/ballerinalang/composer/pull/2013

PR #1574, opened 2017-05-15
fix model undefined error

pahans:


modules/web/js/ballerina/views/ballerina-file-editor.js
```diff
@@ -846,6 +846,11 @@ class BallerinaFileEditor extends BallerinaView {
      * @returns [int]
     */
     getBreakpoints() {
+        // handle if failed to build model
+        if(!this._model) {
```
hemikak: There is a spacing problem here, better to format the code.

pahans: this is fixed in latest master branch

PR #1576, opened 2017-05-15
fix jump to code in compound statements

pahans:


modules/web/js/ballerina/components/compound-statement-decorator.jsx
```diff
@@ -53,6 +53,13 @@ class CompoundStatementDecorator extends React.Component {
         this.setState({innerDropZoneExist: false});
     }
 
+    onJumptoCodeLine() {
```
hemikak: If this method is an overridden method, we need to set {@inheritDoc} here and the super class should have the actual comment. Unless the behavior is changed.

pahans: fixed in https://github.com/ballerinalang/composer/pull/2012

PR #1578, opened 2017-05-16
Improvements and bug fixes

nadeeshaan:
- Add Default else statement
- Increase while statement height when empty,
- Increase block statement width
- Add expression editor to throw statement


modules/web/js/ballerina/ast/statements/if-else-statement.js
```diff
@@ -36,6 +36,8 @@ class IfElseStatement extends Statement {
         this._ifStatement = ifStatement;
 
         this._elseStatement = undefined;
```
NipunaMarcus: we don't need this undefined assignment. No need to specially assign this to a variable.

nadeeshaan: Fixed

PR #1608, opened 2017-05-16
Avoid warnings in the console

roonyh:
* Avoid calling forceUpdate inside the render method of Diagram by calling addChild silently
* Add a default value of empty string for EditableText children prop so input value prop is not set to undefined making it uncontrolled

### DIFF
--- a/modules/web/js/ballerina/ast/resource-definition.js
+++ b/modules/web/js/ballerina/ast/resource-definition.js
@@ -1,3 +1,4 @@
+
 /**
  * Copyright (c) 2016, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
  *
@@ -208,7 +209,7 @@ class ResourceDefinition extends ASTNode {
         let argParamDefHolder = this.findChild(this.getFactory().isArgumentParameterDefinitionHolder);
         if (_.isUndefined(argParamDefHolder)) {
             argParamDefHolder = this.getFactory().createArgumentParameterDefinitionHolder();
-            this.addChild(argParamDefHolder);
+            this.addChild(argParamDefHolder, undefined, true);
         }
         return argParamDefHolder;
     }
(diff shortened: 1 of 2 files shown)


modules/web/js/ballerina/ast/resource-definition.js
```diff
@@ -269,7 +270,7 @@ class ResourceDefinition extends ASTNode {
             if (firstWorkerIndex > -1 && _.isNil(index)) {
                 index = firstWorkerIndex;
             }
-            Object.getPrototypeOf(this.constructor.prototype).addChild.call(this, child, index);
+            Object.getPrototypeOf(this.constructor.prototype).addChild.call(this, child, index, ignoreTreeModifiedEvent);
```
nadeeshaan: seems like this particular line is common to both if and the else clause. If so we can get it out from the clauses

modules/web/js/ballerina/components/editable-text.jsx
```diff
@@ -30,7 +30,7 @@ class EditableText extends React.Component {
     }
 
     renderTextBox() {
-        const {x, y, width, height = 25, onChange, onBlur, onKeyDown, children,
+        const {x, y, width, height = 25, onChange, onBlur, onKeyDown, children="",
```
hemikak: We have to use single quotes here '' . is 'children' here an array ?

roonyh: Single qoutes seems to be added by now in our lint fix process. Children here should be a string. I'll add proptype validations as part of the remaining lint fixes.

modules/web/js/ballerina/ast/resource-definition.js
roonyh: Yeah. good catch! fixed in #1985

PR #1619, opened 2017-05-17
fix lifeline jump to code

pahans:


modules/web/js/ballerina/components/lifeline.jsx
```diff
@@ -28,14 +28,21 @@ import ActionBox from "./action-box";
...
-    
+
     constructor(props){
         super(props);
         let bBox = this.props.bBox;
         this.topBox = new SimpleBBox( bBox.x, bBox.y, bBox.w , lifeLine.head.height );
         this.state = {active: 'hidden'};
     }
 
+    onJumptoCodeLine() {
```
hemikak: Need to have js doc comments.

pahans: fixed in https://github.com/ballerinalang/composer/pull/2012

PR #1630, opened 2017-05-17
Prevent values bleeding between EditableTexts of panel headers

roonyh:
panel title is upadated on enter instead of every key press

### DIFF
--- a/modules/web/js/ballerina/components/editable-text.jsx
+++ b/modules/web/js/ballerina/components/editable-text.jsx
@@ -31,7 +31,7 @@ class EditableText extends React.Component {
 
     renderTextBox() {
         const {x, y, width, height = 25, onChange, onBlur, onKeyDown, children="",
-            className = 'text-input', placeHolder, canUpdate} = this.props;
+            className='text-input', placeholder} = this.props;
         const inputStyle = {
             position: 'absolute',
             top: y - height / 2,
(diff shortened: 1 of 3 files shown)


modules/web/js/ballerina/components/editable-text.jsx
```diff
@@ -31,7 +31,7 @@ class EditableText extends React.Component {
 
     renderTextBox() {
         const {x, y, width, height = 25, onChange, onBlur, onKeyDown, children="",
-            className = 'text-input', placeHolder, canUpdate} = this.props;
+            className='text-input', placeholder} = this.props;
```
hemikak: There should be spacing around the '=' sign.

modules/web/js/ballerina/components/panel-decorator.jsx
```diff
@@ -57,14 +63,26 @@ class PanelDecorator extends React.Component {
...
+    }
+
+    onTitleKeyDown(e) {
+        if(e.keyCode === 13) {
+            const modelType = this.props.model.type.replace('Definition', '');
 
-        // Setter functions take form 'setModelTypeName'. eg: setServiceName
-        this.props.model[`set${modelType}Name`](e.target.value);
+            // Setter functions take form 'setModelTypeName'. eg: setServiceName
+            this.props.model[`set${modelType}Name`](this.state.editingTitle);
```
hemikak: A bit risky to use a template to find and all a function. Can we pass the setter method as a prop?

modules/web/js/ballerina/components/utils/tag-component.jsx
```diff
@@ -68,14 +68,18 @@ class TagController extends React.Component {
      * */
     onKeyDown(e) {
         if (e.keyCode === 13) {
+            let validate = this.props.validateInput;
+            let variableDeclaration = this.state.editValue.replace("=", "").replace(";", "");
```
hemikak: We need to use single quotes here.

modules/web/js/ballerina/components/panel-decorator.jsx
sumuditha-viraj: Agree with Hemika.

PR #1635, opened 2017-05-17
Service / Connector level variables are foldable

roonyh:


modules/web/js/ballerina/visitors/sizing-utils.js
```diff
@@ -591,6 +591,11 @@ class SizingUtil {
 
     getConnectorLevelVariablesHeight(node) {
         let height = 65;
+
+        if(!node.viewState.variablesExpanded){
+            return 35;
```
hemikak: Should this go to designer-defaults ?

PR #1676, opened 2017-05-23
Stop rendering of text box making the text appear to jump

roonyh:
previous: 
peek 2017-05-23 11-44

now: 
peek 2017-05-23 11-46

### DIFF
--- a/modules/web/js/ballerina/components/editable-text.jsx
+++ b/modules/web/js/ballerina/components/editable-text.jsx
@@ -26,25 +26,38 @@ class EditableText extends React.Component {
             editing: false
         };
 
+        const textBoxPadding = 3;
+        const textBoxBorderSz = 2;
+
+        this.inputStyles = {
+            position: 'absolute',
+            paddingLeft: textBoxPadding,
+            borderWidth: textBoxBorderSz
+        }
+
         this.renderTextBox = this.renderTextBox.bind(this);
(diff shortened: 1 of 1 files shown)


modules/web/js/ballerina/components/editable-text.jsx
```diff
@@ -26,33 +26,46 @@ class EditableText extends React.Component {
             editing: false
         };
 
+        const textBoxPadding = 3;
```
NipunaMarcus: can't we move this to default config?

roonyh: I did thought about it. Should we move all the small styling configs there? There are some stuff that definitely should be configurable so should be defined in the config file. But moving every small thing there should make the config unnecessarily huge imo.

hevayo: IMO this should be captured in CSS, if we make the default too granular it will be difficult to maintain.

roonyh: Thing is this value affects the svg `<text>` element as well. the `x` value of it should be added the paddingLeft and and borderWith properties so it does not appear as if it jumps. See https://github.com/ballerinalang/composer/pull/1676/files#diff-905b83fb1d50e077c850ce881d193e1eR103

PR #1677, opened 2017-05-23
Delete unused code & Add webpack plugin to inspect unused js files

kaviththiranga:


modules/web/webpack.config.js
```diff
@@ -85,7 +86,8 @@ var config = {
     },
     plugins: [
         extractCSSBundle,
-        extractThemes
+        extractThemes,
+        new UnusedFilesWebpackPlugin({pattern: '**/*.js'})
```
roonyh: Can we capture `.jsx` files as well? how about `**/*.*`?

kaviththiranga: *.* will complain about unused images and css too right? can we guarantee there is no css or svg which is not included using webpack aware api

kaviththiranga: WDYT?

kaviththiranga: or may be we can set it as `js/**/*.*`

roonyh: Ah. yeah. `'**/*.js*'` should capture jsx as well ne?

kaviththiranga:
@roonyh Updated. Please check

roonyh:
LGTM :+1:

PR #1680, opened 2017-05-23
Add tryServiceURL config to workspace-service-config.yaml

sumuditha-viraj:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/launcher/LauncherConstants.java
```diff
@@ -55,6 +55,11 @@
     public static final String EXECUTION_TERMINATED = "EXECUTION_TERMINATED";
     public static final String SET_BAL_PATH_MESSAGE = "Please set BALLERINA_HOME environment variable pointing to " +
                                                       "ballerina runtime.";
+
+    public static final String SERVER_CONNECTOR_STARTED_AT_HTTP_DEFAULT_PORT_LOG = "ballerina: started server " +
+                                                                                    "connector http-9090";
+    public static final String SERVER_CONNECTOR_STARTED_LOG = "ballerina: started server connector";
+    public static final String DEFAULT_STARTED_SERVICE_URL = "http://localhost:9090";
```
pahans: can we guarantee that service will run on port 9090?

sumuditha-viraj: In cloud scenario we have to guarantee that service is started on a certain port because we have to configure that in the proxy. Due to that reason, if we cannot guarantee that the service is started on a certain port, in cloud scenarios its anyway broken. (With the current implementation). Due to that, I have added 9090 which is the default assuming service is always starting on a certain port.  BTW, I will remove this so that it will not change the previous behaviour for on premise use case.  For cloud scenario we have to make that assumption anyway.

sumuditha-viraj: Updated the PR with required changes.

PR #1681, opened 2017-05-23
Unsubscribe from dragDrop and messageManager events when a StatementComponent unmounts

roonyh:
Otherwise when a component is unmounted for example by deleting, those event handlers will still fire. This results in warining 'setState is called on an unmounted component'.

### DIFF
--- a/modules/web/js/ballerina/components/statement-container.jsx
+++ b/modules/web/js/ballerina/components/statement-container.jsx
@@ -29,10 +29,8 @@ class StatementContainer extends React.Component {
     constructor(props, context) {
         super(props, context);
         const {dragDropManager, messageManager} = context;
-        dragDropManager.on('drag-start', this.startDropZones.bind(this));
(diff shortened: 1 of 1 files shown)


modules/web/js/ballerina/components/statement-container.jsx
```diff
@@ -29,17 +29,30 @@ class StatementContainer extends React.Component {
...
             statementDropZoneActivated: false,
             dropZoneDropNotAllowed: false
         }
     }
 
+    componentDidMount() {
+        const { dragDropManager, messageManager } = this.context;
+        dragDropManager.on('drag-start', this.startDropZones);
+        dragDropManager.on('drag-stop', this.stopDragZones);
+		messageManager.on('message-draw-start', this.stopDragZones);
```
sumuditha-viraj: Please fix formatting issues.

roonyh: Fixed the lint errors including this one in #1990

PR #1686, opened 2017-05-26
Avoid adding empty else block when no else block present

### LINKED ISSUE #565: Empty else block is added if there is no else block
Steps: Deserialize and serialize following source code

`
function sum(int a)(int ) {
	int x;
	if (a > 0) {
		x = sum(a - 1);
		a = a + x;		
	}
	return a;	
}
`
gives following source code

`
function sum(int a)(int ) {
	int x;
	if (a > 0) {
		x = sum(a - 1);
		a = a + x;		
	}
	else {
		}return a;	
}
`

roonyh:
Fixes #565

### DIFF
--- a/modules/web/js/ballerina/ast/statements/if-else-statement.js
+++ b/modules/web/js/ballerina/ast/statements/if-else-statement.js
@@ -36,9 +36,6 @@ class IfElseStatement extends Statement {
         this.addChild(ifStatement);
         this._ifStatement = ifStatement;
 
-        this._elseStatement = new ElseStatement();
-        this.addChild(this._elseStatement);
-
         this._elseIfStatements = [];
         this.type = "IfElseStatement";
     }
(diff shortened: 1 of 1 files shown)


modules/web/js/ballerina/ast/statements/if-else-statement.js
```diff
@@ -156,6 +153,12 @@ class IfElseStatement extends Statement {
             });
         });
 
+        if(jsonNode.else_statement) {
+            const elseStatement = new ElseStatement();
```
hemikak: Should we use the ballerina AST factory to create an ElseStatement ?

roonyh: Yeah. Changed to use BAST factory in https://github.com/ballerinalang/composer/pull/1980

PR #1709, opened 2017-05-30
Fix binary and array init issue

nadeeshaan:


modules/web/js/ballerina/ast/expressions/array-init-expression.js
```diff
@@ -45,7 +45,12 @@ class ArrayInitExpression extends Expression {
     }
 
     generateExpression() {
-        this._expression = this.getVariableName();
+        var generatedExpression = '[';
+        _.each(this.getChildren(), function (child) {
+            generatedExpression += child.getExpression() + ",";
```
hemikak: We need to avoid using double quotes with the linter rules.

PR #1711, opened 2017-05-30
Fix the dimension issues in annotation definition view.

NipunaMarcus:


modules/web/js/ballerina/components/annotation-attribute-decorator.jsx
```diff
@@ -145,12 +144,18 @@ class AnnotationAttributeDecorator extends React.Component {
         return (
             <g className="attribute-content-operations-wrapper">
                 <g onClick={() => this.onClickVariableTextBox()}>
-                    <rect x={bBox.x + 50} y={bBox.y + 50} width={330} height={30}
+                    <rect x={bBox.x + DesignerDefaults.panel.body.padding.left}
```
hevayo: Shall we do the bBox calculation above the return so the code would be easy to read .

PR #1715, opened 2017-05-31
Load packages in program directory

sumuditha-viraj:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/Constants.java
```diff
@@ -42,6 +42,8 @@
     public static final String SYS_BAL_COMPOSER_HOME = "bal.composer.home";
 
     public static final String COMPOSER_HOME_NOT_FOUND_ERROR_MESSAGE = "Error: Unable to find ballerina composer home.";
+
+    public static final int DIRECTORY_DEPTH = 30;
```
hevayo: Shall we make this like 10 ?

sumuditha-viraj: Fixed this.

PR #1773, opened 2017-06-02
Updating font-wso2

srsharon:


modules/web/js/ballerina/item-provider/initial-definitions.js
```diff
@@ -254,17 +254,17 @@ var createThrowStatementTool = {
 var createAbortStatementTool = {
     id: "Abort",
     name: "Abort",
-    iconSrc: require("./../../../images/tool-icons/throw.svg"),
-    cssClass: "icon fw fw-throw",
+    iconSrc: require("./../../../images/tool-icons/abort.svg"),
```
jeradrutnam: This will be removed later on. As we are going to remove all the svg file imports. And use Font WSO2 icons.json in the graph and Font Icons in HTML. So its not really necessary.

modules/web/js/ballerina/item-provider/initial-definitions.js
```diff
@@ -254,17 +254,17 @@ var createThrowStatementTool = {
...
     title: "Abort",
     nodeFactoryMethod: DefaultBallerinaASTFactory.createAbortStatement
 };
 
 var createTransactionAbortedStatementTool = {
     id: "Transaction",
     name: "Transaction",
-    iconSrc: require("./../../../images/tool-icons/try-catch.svg"),
-    cssClass: "icon fw fw-try-catch",
+    iconSrc: require("./../../../images/tool-icons/transaction.svg"),
```
jeradrutnam: Same here. See above comment.

srsharon:
Yes agreed. We have both wso2-font & the SVG options in the composer code. We will have to clean the code with the font where its required.  +1 for removing SVG and maintaining the icons with font-wso2.

PR #1788, opened 2017-06-06
Added replace plugin to update version on build.

hemikak:


modules/web/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "ballerina-composer",
-  "version": "0.8.1",
+  "version": "0.88.0",
```
pahans: 0.88.0 or 0.8.8 ?

PR #1810, opened 2017-06-07
 Timeout ui improvements.

manuranga:

### DIFF
--- a/modules/web/js/ballerina/configs/designer-defaults.js
+++ b/modules/web/js/ballerina/configs/designer-defaults.js
@@ -128,7 +128,7 @@ export const blockStatement = {
         width: 50,
         height: statement.height,
         paramPaddingX: 5,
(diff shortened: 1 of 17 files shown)


modules/web/js/ballerina/configs/designer-defaults.js
```diff
@@ -263,4 +264,11 @@ export const panelHeading = {
         heading: 14,
         padding: 3
     }
-};
\ No newline at end of file
+};
+
+export const timeout = {
```
hevayo: Can this be renamed to a more generic name ? then we can use the same config in another statement type if required.

manuranga: generic values are already defined in blockStatement, this values are specific to the the timeout, because work "Timeout" is longer.

Issue #1771, opened 2017-06-02
Error while debugging remotely in Cloud Composer


SabraO:
![test2](https://cloud.githubusercontent.com/assets/5234592/26727178/658636e4-47c3-11e7-929b-8b3066c4062d.png)

The above error occurs when we provide the URL of the debug service for the Composer container in Cloud.

pahans:
Hi, @SabraO 
Does Debug Service button works on this scenario?

PR #1824, opened 2017-06-09
fix remote debugging ws url

pahans:


pahans:
related to https://github.com/ballerinalang/composer/issues/1771

PR #1834, opened 2017-06-09
Timeout UI Improvements.

manuranga:


manuranga:
same as #1810 , but against ballerina-virtual-machine.

PR #1917, opened 2017-06-12
Made annotation attributes remove button display on hover

jeradrutnam:


modules/web/css/ballerina-file-editor.css
```diff
@@ -892,7 +892,7 @@ input#package-name-input {
 }
 
 .annotation-table td{
-    height: 20px;
+    height: 18px;
```
hemikak: Having height 18px might cause an issue with calculating the height of the container as the default height is taken as 20px. Have to update here as well https://github.com/ballerinalang/composer/blob/master/modules/web/js/ballerina/visitors/sizing-utils.js#L595 .

jeradrutnam: Shall we change it there as well? And I tested after changing it, but didn't overcome any issue.

jeradrutnam: Reverted this and made a different change to overcome this

jeradrutnam:
Changes moved to #1936

PR #1978, opened 2017-06-15
Transform bug fixes, UX improvements and Lint fixes

keizer619:


modules/web/js/ballerina/ast/node.js
```diff
@@ -271,6 +271,21 @@ class ASTNode extends EventChannel {
     }
 
     /**
+     * Filter matching children belonging to current scope from the predicate function
+     * @param predicateFunction a function returning a boolean to match filter condition from children
+     * @return {[ASTNode]} array of matching AST nodes
```
hemikak: Should be as follows: 

@param {func} a function returning a boolean to match filter condition from children.
@return {ASTNode[]} array of matching AST nodes.

modules/web/js/ballerina/components/transform-render.js
```diff
@@ -977,7 +959,36 @@ class TransformRender {
             alwaysRespectStubs: true,
         }];
     }
-}
 
 
+    onRemove(id, func, removeFunction, reference) {
```
hemikak: Will need jsDocs here.

modules/web/js/ballerina/components/transform-render.js
```diff
@@ -977,7 +959,36 @@ class TransformRender {
...
+        const self = this;
```
hemikak: Since we are using arrow functions here, we should be able to get rid of 'self' const.

modules/web/js/ballerina/components/transform-render.js
```diff
@@ -576,13 +582,14 @@ class TransformRender {
 
         if (type == 'source') {
             subPlaceHolder = 'leftType';
-            $(`#${this.placeHolderName}`).find('.leftType').append(newVar);
-            this.addSource(property, this, true);
+            $('#' + this.placeHolderName).find('.leftType').append(newVar);
+            this.addSource(property, this, false);
```
hemikak: Was this changed by purpose, or did the linter changed it ?

modules/web/js/ballerina/components/transform-statement-decorator.jsx
```diff
@@ -769,39 +756,63 @@ class TransformStatementDecorator extends React.Component {
...
+        if (_.isUndefined(targetSelection)){
+            alerts.error('Mapping target "' + currentSelection + '" cannot be found');
             return false;
         }
+
+        var removeFunc = function(id) {
+            self.mapper.removeType(id);
+            _.remove(self.props.model.getOutput(), function(currentObject) {
+                                                                return currentObject.getVariableName() === id;
+                                                            })
```
hemikak: Is there an indentation issue here ? , I think we can use arrow function here as well.

modules/web/js/ballerina/components/transform-statement-decorator.jsx
```diff
@@ -769,39 +756,63 @@ class TransformStatementDecorator extends React.Component {
...
-        let targetSelection = _.find(predefinedStructs, { name: currentSelection });
-        if (_.isUndefined(targetSelection)) {
-            alerts.error(`Mapping target "${  currentSelection  }" cannot be found`);
+        var targetSelection = _.find(predefinedStructs, { name: currentSelection});
+        if (_.isUndefined(targetSelection)){
+            alerts.error('Mapping target "' + currentSelection + '" cannot be found');
             return false;
         }
+
+        var removeFunc = function(id) {
```
hemikak: You can use const here.

modules/web/js/ballerina/components/transform-render.js
keizer619: @hemikak : true -> false is purposely done and same line has been refactored by linter too.

hemikak: 👍

PR #1981, opened 2017-06-15
Fix the issue where attachment cannot be added to annotation definition

NipunaMarcus:


modules/web/js/ballerina/components/utils/tag-component.jsx
```diff
@@ -122,9 +152,10 @@ class TagController extends React.Component {
               <rect
                 x={componentData.components.openingBracket.x}
                 y={componentData.components.openingBracket.y}
-                width={componentData.components.closingBracket.x - (componentData.components.openingBracket.x - 3)
-                      + componentData.components.closingBracket.w}
-                height={25} className={this.props.groupClass}
+                width={componentData.components.closingBracket.w +
```
manuranga: Better to move to a const out of the XML part.

PR #1982, opened 2017-06-15
Added linter fixes for source-gen and swagger-json visitor folders.

hemikak:


modules/web/.eslintrc.js
```diff
@@ -21,7 +21,9 @@ module.exports = {
                 "ClassDeclaration": true
             }
         }],
-        "valid-jsdoc": "warn",
+        "valid-jsdoc": ["warn", {
```
pahans: shall we make this off?

modules/web/js/ballerina/visitors/source-gen/statement-visitor-factory.js
```diff
@@ -107,6 +107,8 @@ class StatementVisitorFactor {
         } else if (ASTFactory.isCommittedStatement(statement)) {
             return new CommittedStatementVisitor(parent);
         }
+
+        return undefined;
```
manuranga: isn't just `return` better ?

Issue #1920, opened 2017-06-12
Unable to rename an endpoint


srsharon:
Once the endpoint name is changed it will not reflect on the endpoint unless you switch to code view and come back to design view 

![image](https://user-images.githubusercontent.com/10986966/27021082-aa905b32-4f63-11e7-8a5b-4c69d5a73a84.png)

nadeeshaan:
Closing since the fix is merged

PR #1994, opened 2017-06-15
Fix Issue No 1920

nadeeshaan:
Resolves Issue #1920


modules/web/webpack.config.js
```diff
@@ -5,7 +5,7 @@ const UnusedFilesWebpackPlugin = require('unused-files-webpack-plugin').UnusedFi
 
 const extractThemes = new ExtractTextPlugin('./[name].css');
 const extractCSSBundle = new ExtractTextPlugin('./bundle.css');
-let exportConfig = {};
+var exportConfig = {};
```
hemikak: Can we use let here ?

nadeeshaan: If we do so, we won't be able to use the **npm run dev** command

nadeeshaan:
Resolves #1920

PR #2020, opened 2017-06-19
New completer for source view

nadeeshaan:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/util/completion/PackageItemResolver.java
```diff
@@ -0,0 +1,43 @@
...
+     */
+    private PackageItemResolver(){
+    }
+
+    /**
+     * Static get instance method
+     * @return PackageItemResolver - PackageItemResolver instance
+     */
+    public static PackageItemResolver getInstance(){
+        if(instance==null){
```
hemikak: Spacing issue here around the brackets.

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/util/completion/PackageItemResolver.java
```diff
@@ -0,0 +1,43 @@
...
+            synchronized (mutex){
+                if(instance==null) instance= new PackageItemResolver();
```
hemikak: Shall we avoid single line if bodies ?

PR #2024, opened 2017-06-19
Eslint and doc fixes

manuranga:


modules/web/js/ballerina/components/block-statement-decorator.jsx
```diff
@@ -63,22 +85,26 @@ class BlockStatementDecorator extends React.Component {
...
-        $(container).find('.view-source-btn').trigger('click');
+    onJumpToCodeLine() {
+        document.getElementsByClassName('view-source-btn')[0].click();
     }
 
-    onUpdate(text) {
+    /**
+     * Call-back for when a new value is entered via expression editor.
+     */
+    onUpdate() {
```
hemikak: Is this an unused method ? Body is empty.

manuranga: Ya it's not used yet but i though @kaviththiranga was planing to use it to validate later, I'll remove it for now then.

manuranga: actually I added an TODO for now since we are going to need this in the future.

PR #2031, opened 2017-06-20
Adding the initial function resolving through language server

nadeeshaan:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/consts/LangServerConstants.java
```diff
@@ -80,6 +80,12 @@
     public static final String PACKAGES = "packages";
 
     /**
+     * Resolving Cafeterias
```
ramith: is this intentional ? 'Cafeterias' ?

nadeeshaan: @ramith no, this is a mistake in the current PR. This has been refactored

Issue #2029, opened 2017-06-20
[Feature Request] Ability to import a Swagger JSON into the composer 


handakumbura:
At the moment we have a similar capability in the run-time. 

In the composer, we can continue API definition in the embedded Swagger editor once a service skeleton has been created. Since the skeleton creation is as simple as dragging in a service construct from the design palette this behavior is acceptable but it would be even more convenient if we can provide an option to import a swagger definition and based on a few predefined field values a skeleton can be generated.

PR #2118, opened 2017-06-23
Added feature to create a service using a swagger definition file. Fixed minor bugs and updated service.

hemikak:
Fixes: https://github.com/ballerinalang/composer/issues/2029

### DIFF
--- a/modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/Workspace.java
+++ b/modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/Workspace.java
@@ -33,7 +33,14 @@ public interface Workspace {
     
     JsonArray listDirectoriesInPath(String path) throws IOException;
     
-    JsonArray listFilesInPath(String path) throws IOException;
+    /**
+     * Returns a JSON array with the files in a specific path.
+     * @param path The path.
+     * @param extensions A list of extensions to support.
+     * @return Json array of files.
+     * @throws IOException
+     */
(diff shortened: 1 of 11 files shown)


handakumbura:
It would be better if we can put in **ballerina.net.http** package import statement for the service that gets generated.

handakumbura:
@hemikak the composer gets stuck in the swagger editor mode for the following definition. looks like the issue is with the @swagger:ParametersInfo {}. Can we smarten up the logic to not put this tag in if the swagger parameters definition is one that is not supported in our current implementation? 

```
{
  "swagger": "2.0",
  "basePath": "/branchlocator",
  "info": {
    "version": "1.0.0",
    "title": "Sample Service"
  },
  "paths": {
    "/product": {
      "post": {
        "operationId": "product",
        "responses": {
          "default": {
            "description": "Default Response"
          },
          "405": {
            "description": "Invalid input"
          }
        },
        "summary": "This resource returns the banks name and code based on the zip code provided.",
        "consumes": [
          "application/json"
        ],
        "produces": [
          "application/json"
        ],
        "parameters": [
          {
            "in": "body",
            "name": "body",
            "description": "The branch locator payload containing the zip code of the consumer.",
            "required": true,
            "schema": {
              "$ref": "#/definitions/RequestPayload"
            }
          }
        ]
      }
    }
  },
  "definitions": {
    "RequestPayload": {
      "type": "object",
      "properties": {
        "BranchLocator": {
          "$ref": "#/definitions/ZipCode"
        }
      }
    },
    "ZipCode": {
      "type": "object",
      "properties": {
        "ZipCode": {
          "type": "string"
        }
      }
    }
  }
}
```

Issue #583, opened 2017-02-14
Files not displayed on Open File modal when saved on root directory. 


erangatl:
- Save file in the root directory (D:\)
- File is not displayed in the Open File tree view  

![image](https://cloud.githubusercontent.com/assets/20179540/22915997/3dd7f3e8-f2a1-11e6-8f49-d39fddf89034.png)

kaviththiranga:
Fixed with https://github.com/ballerinalang/composer/pull/2122

PR #2122, opened 2017-06-25
List bal files in file-system roots & 	Filter out windows system files

kaviththiranga:


kaviththiranga:
Includes fix for https://github.com/ballerinalang/composer/issues/583

Issue #2150, opened 2017-06-28
Annotation support for Consumes and Produces annotations.


hemikak:
This is not yet supported by the core. Hence will be ignored while switching from swagger view.

hemikak:
Solved with https://github.com/ballerinalang/ballerina/blob/master/modules/ballerina-native/src/main/ballerina/ballerina/net/http/annotation.bal#L41

PR #2198, opened 2017-06-30
Disabled consumes and produces support for annotations.

hemikak:


hemikak:
Related to https://github.com/ballerinalang/composer/issues/2150

Issue #2263, opened 2017-07-04
Support var keyword


maheeka:
var x, b = testFunction()

above syntax must be supported

PR #2270, opened 2017-07-04
add var keyword for assignment statements

maheeka:
https://github.com/ballerinalang/composer/issues/2263


maheeka:
Depends on https://github.com/ballerinalang/ballerina/pull/2844

PR #2295, opened 2017-07-07
Accumulator fixes

sumuditha-viraj:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/util/resolvers/AnnotationAttachmentResolver.java
```diff
@@ -0,0 +1,93 @@
+package org.ballerinalang.composer.service.workspace.langserver.util.resolvers;
```
maheeka: License header missing

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/util/resolvers/PackageActionsAndFunctionsResolver.java
```diff
@@ -0,0 +1,26 @@
+package org.ballerinalang.composer.service.workspace.langserver.util.resolvers;
```
maheeka: License header

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/util/resolvers/AnnotationAttachmentResolver.java
sumuditha-viraj: Fixed in #2300

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/util/resolvers/PackageActionsAndFunctionsResolver.java
sumuditha-viraj: Fixed in #2300

PR #2306, opened 2017-07-07
Adding continue statement.

manuranga:


manuranga:
Known issues: position doesn't get updated when new items are inserted above the continue.

manuranga:
Fixed above issue.

Issue #2329, opened 2017-07-11
Cannot add struct string fields without default values


maheeka:
Default value "" always get added for string types when adding a field. Once the default value is added, there is no way of removing it as well. 

<img width="638" alt="image" src="https://user-images.githubusercontent.com/1029806/28069913-cb2ff782-6668-11e7-9207-471a80f64bdc.png">

PR #2334, opened 2017-07-12
Handling the default values of the struct string fields

NatashaWso2:
Fixed : https://github.com/ballerinalang/composer/issues/2329

### DIFF
--- a/modules/web/js/ballerina/components/struct-definition.jsx
+++ b/modules/web/js/ballerina/components/struct-definition.jsx
@@ -83,11 +83,10 @@ class StructDefinition extends React.Component {
      */
     addQuotesToString(dataType, value) {
         if (dataType === 'string' && !/".*?"$/.test(value)) {
-            if (value !== '') {
-                return `"${value}"`;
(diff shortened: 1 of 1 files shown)


modules/web/js/ballerina/components/struct-definition.jsx
```diff
@@ -76,9 +76,18 @@ class StructDefinition extends React.Component {
...
+     * Handle the default value if the data type is a string
+     * @param {string} dataType - Data type of the new struct
+     * @param {string} value - Value of the new identifier
+     */
     addQuotesToString(dataType, value) {
         if (dataType === 'string' && !/".*?"$/.test(value)) {
-            return `"${value}"`;
+            if (value !== '') {
+                return `"${value}"`;
+            } else if (value === '') {
```
maheeka: this is the negate of the if condition. we can just have if-else. Better to have the empty check first though for readability.

NatashaWso2: Fixed with commit : 9aba25a

PR #2355, opened 2017-07-13
When switching back from source mode if transform expanded view is active its updated

roonyh:


modules/web/js/ballerina/views/design-view.jsx
```diff
@@ -31,6 +31,10 @@ class DesignView extends React.Component {
         this.diagramContainer = ref;
     }
 
+    getTransformActive() {
```
nadeeshaan: Shall we add the doc comments here?

roonyh: added doc comments for the transform active state getter and setter

PR #2388, opened 2017-07-17
Global constant definitions rhs can be an expression

### LINKED ISSUE #2296: Constant definition (Global) view is not working with new const def changes

roonyh:
We cant visit the included variable def statement as a whole because that results in a recursive loop. (because variableDef is pointed to the parent constDef)
So visiting only the rhs. Identifier and the type are asigned to the json object seperately.
Fixes #2296


roonyh:
Fixes #2296

Issue #2326, opened 2017-07-11
Auto generated variable names are different in functions and resources


maheeka:
Example: Add an action invocation in a resource and a function

<img width="393" alt="image" src="https://user-images.githubusercontent.com/1029806/28067739-1f3d0e04-6660-11e7-9787-98b987abb699.png">

<img width="503" alt="image" src="https://user-images.githubusercontent.com/1029806/28067727-12ba5fd8-6660-11e7-8140-c363e6fe3b18.png">

i1,i2 variable names must be removed in resources. Approach followed in functions is correct. Same is applicable for variable def statements.

Issue #2402, opened 2017-07-17
Issues when adding imports


dilinisg:
Pack - tools (14/7)

**Steps**
1. Select an import from the drop down list
2. Click on the canvas without pressing Enter 

**Issues**
It does not show that the import is added if we don't press enter after selecting the import from the drop down list. But variables cannot be added without pressing enter. 

When you click on the text field again the behaviour is as follows.

![imp](https://user-images.githubusercontent.com/1845370/28267534-1e52f082-6b18-11e7-92f2-f4f99b8cee93.png)

PR #2432, opened 2017-07-19
Making auto generated variable names consistent and fixing the issue when adding imports

NatashaWso2:
This fixes issues: https://github.com/ballerinalang/composer/issues/2326 and https://github.com/ballerinalang/composer/issues/2402

### DIFF
--- a/modules/web/js/ballerina/ast/worker-declaration.js
+++ b/modules/web/js/ballerina/ast/worker-declaration.js
@@ -158,7 +158,7 @@ class WorkerDeclaration extends ASTNode {
         CommonUtils.generateUniqueIdentifier({
             node: this,
             attributes: [{
-                defaultValue: 'Worker',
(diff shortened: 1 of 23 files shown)


modules/web/js/ballerina/ast/worker-declaration.js
```diff
@@ -158,7 +158,7 @@ class WorkerDeclaration extends ASTNode {
         CommonUtils.generateUniqueIdentifier({
             node: this,
             attributes: [{
-                defaultValue: 'newWorker',
+                defaultValue: 'Worker',
```
hevayo: if these are identifiers shouldn't we start with simple ?

NatashaWso2: Fixed in commit :  fd29ad3

Issue #2439, opened 2017-07-19
Continue statement can be added anywhere in a lifeline


dilinisg:
pack - 19/07

Continue statement can be added anywhere in a lifeline

NipunaMarcus:
This issue is fixed hence closing.

PR #2467, opened 2017-07-19
Validating where the continue statement can be added in the lifeline

NatashaWso2:
This fixes the issue: https://github.com/ballerinalang/composer/issues/2439

### DIFF
--- a/modules/web/js/ballerina/ast/statements/continue-statement.js
+++ b/modules/web/js/ballerina/ast/statements/continue-statement.js
@@ -51,6 +51,16 @@ class ContinueStatement extends Statement {
     getStatementString() {
         return 'continue';
     }
+
+    /**
+     * Define what type of nodes that this node can be added as a child.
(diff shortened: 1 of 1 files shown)


modules/web/js/ballerina/ast/statements/continue-statement.js
```diff
@@ -51,6 +51,16 @@ class ContinueStatement extends Statement {
...
         return 'continue';
     }
+
+    /**
+     * Define what type of nodes that this node can be added as a child.
+     * @param {ASTNode} node - Parent node that this node becoming a child of.
+     * @return {boolean} true|false.
+     * */
+    canBeAChildOf(node) {
+        return this.getFactory().isWhileStatement(node) || this.getFactory().isIfElseStatement(node) ||
```
hemikak: This way, we can add continue statements to if conditions too.

NatashaWso2: Fixed in commit 3421e1e

Issue #2410, opened 2017-07-18
Connector declaration formatting issues


dilinisg:
![varcon](https://user-images.githubusercontent.com/1845370/28301675-1422e7b4-6ba6-11e7-9b48-2433b5d9caf6.png)

Issue #2327, opened 2017-07-11
Annotation view filter should do a contains search


maheeka:
Should have the same behaviour as import package search. Search is currently _startsWith_. Should change to _contains_. 

<img width="279" alt="image" src="https://user-images.githubusercontent.com/1029806/28068140-d001ca26-6661-11e7-9125-0a9dc9267807.png">
<img width="279" alt="image" src="https://user-images.githubusercontent.com/1029806/28068152-d8331a1a-6661-11e7-8816-31bfa5d56a5d.png">

vs

<img width="381" alt="image" src="https://user-images.githubusercontent.com/1029806/28068158-e1654ce8-6661-11e7-8270-da7769ec849b.png">

PR #2475, opened 2017-07-20
Fixing connector declaration formatting issues and the annotator filter view

NatashaWso2:
This fixes issues: https://github.com/ballerinalang/composer/issues/2410 and https://github.com/ballerinalang/composer/issues/2327


modules/web/js/ballerina/components/global-item.jsx
```diff
@@ -91,8 +91,19 @@ export default class GlobalDefinitionItem extends React.Component {
             className = 'global-definition-item-hightlighted';
         }
 
+        const maxLengthText = 38;
+        let globalItemValue = this.props.getValue(this.props.globalDec);
+
+        if (globalItemValue.length > maxLengthText) {
+            globalItemValue = ((globalItemValue.substring(0, maxLengthText - 3)) + '...');
+        }
```
hemikak: Can use https://lodash.com/docs/4.17.4#truncate  here.

modules/web/js/ballerina/components/annotation-container.jsx
```diff
@@ -367,8 +374,22 @@ class AnnotationContainer extends React.Component {
...
+            if (i < parts.length - 1) {
+                highlightedString.push(<span key={i + 100}><b>{value}</b></span>);
+            }
+        });
+
         return (
-            <span>{suggestion.name}</span>
+            <div>
+                {highlightedString}
+            </div>
```
hemikak: Can we send a <span></span> instead of div ?

modules/web/js/ballerina/components/annotation-container.jsx
```diff
@@ -301,9 +302,9 @@ class AnnotationContainer extends React.Component {
      * @memberof AnnotationContainer
      */
     getAnnotationPackageNameSuggestions(value) {
+        const matches = [];
         const escapedValue = this.escapeRegexCharacters(value.trim());
-
-        const regex = new RegExp(`^${escapedValue}`, 'i');
+        const regex = new RegExp(escapedValue, 'i');
```
hemikak: Is this regex used to check a certain string contains in another string ?
If so, can use https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/includes .

Example: https://github.com/ballerinalang/composer/blob/master/modules/web/js/ballerina/components/utils/autosuggest-html.jsx#L156

NatashaWso2: Fixed with commit  f423d19. Used the includes method

modules/web/js/ballerina/components/annotation-container.jsx
NatashaWso2: Fixed with commit f423d19. Using a span will not break others

modules/web/js/ballerina/components/global-item.jsx
NatashaWso2: Fixed with commit f423d19

modules/web/js/ballerina/components/annotation-container.jsx
```diff
@@ -314,7 +314,13 @@ class AnnotationContainer extends React.Component {
             };
         });
 
-        return packageNameSuggestions.filter(packageNameObject => regex.test(packageNameObject.name));
+        packageNameSuggestions.forEach((sug) => {
+            if ((sug.name).includes(escapedValue)) {
+                matches.push(sug);
+            }
+        });
```
hemikak: We can use the filter function here right ? instead for forEach and adding it to an array.

return packageNameSuggestions.filter((sug) => { return sug.name.includes(escapedValue); }));

NatashaWso2: Fixed with 85f1706

PR #2486, opened 2017-07-20
fix transform issues

keizer619:
1. Allow to map "any" type in the transform
2. Recursion same type as a property type

### DIFF
--- a/modules/web/js/ballerina/components/transform-expanded.jsx
+++ b/modules/web/js/ballerina/components/transform-expanded.jsx
@@ -577,7 +577,7 @@ class TransformExpanded extends React.Component {
             property.packageName = field.getPackageName();
 
             let innerStruct = this.getStructDefinition(property.packageName, property.type);
-            if (innerStruct != null) {
+            if (innerStruct != null && typeName != property.type) {
                 property.innerType = this.createType(property.name, property.type, innerStruct, true);
             }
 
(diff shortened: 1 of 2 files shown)


modules/web/js/ballerina/components/transform-expanded.jsx
```diff
@@ -577,7 +577,7 @@ class TransformExpanded extends React.Component {
             property.packageName = field.getPackageName();
 
             let innerStruct = this.getStructDefinition(property.packageName, property.type);
-            if (innerStruct != null) {
+            if (innerStruct != null && typeName != property.type) {
```
hemikak: !==

modules/web/js/ballerina/components/transform-render.js
```diff
@@ -784,9 +784,10 @@ addComplexParameter(parentId, struct) {
             maxConnections: 1,
             anchor: ['Left'],
             beforeDrop(params) {
-            // Checks property types are equal
-                const isValidTypes = self.getPropertyType(params.sourceId).toLowerCase()
-                                        == self.getPropertyType(params.targetId).toLowerCase();
+                // Checks property types are equal or type is any
+                const sourceType = self.getPropertyType(params.sourceId).toLowerCase();
+                const targetType = self.getPropertyType(params.targetId).toLowerCase()
+                const isValidTypes = sourceType == targetType || sourceType == "any" || targetType == "any"
```
hemikak: Better to use ===

PR #2513, opened 2017-07-20
Improvements for fork join suggestions completion

nadeeshaan:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/CompletionItemAccumulator.java
```diff
@@ -1055,7 +1055,9 @@ public void visit(ForkJoinStmt forkJoinStmt) {
             parameter.accept(this);
             join.define(parameter.getSymbolName(), parameter);
 
-            if (!(parameter.getType() instanceof BMapType)) {
+            // TODO: Join statement's parameter type is null when the parameter name is map. For the moment, instead of
+            // type check we check the type name
+            if (!(parameter.getTypeName().getName().equals("map"))) {
```
sumuditha-viraj: We do not need to throw Semantic Exceptions in CompletionItemAccumulator.java. Instead of altering the "condition" we can remove following "throw new SemanticException()"

nadeeshaan: In the sense this whole condition block invalidates and we can remove that.

sumuditha-viraj: +1

PR #2562, opened 2017-07-21
fix applying viewId instead of placeholderName

keizer619:


modules/web/js/ballerina/components/transform-render.js
```diff
@@ -968,7 +968,7 @@ addComplexParameter(parentId, struct) {
         const xTargetPointer = 0;
         let yTargetPointer = 0;
         const functionGap = 30;
-        const svgLines = $('#' + self.placeHolderName + '> svg');
+        const svgLines = $('#' + self.viewId + '> svg');
```
pahans: you can use a es6 string template string here. ``` $(`#${self.viewId}> svg`);```

hemikak: I think @roonyh suggested to use 'this.container.find' instead of $

roonyh: yep

roonyh: We have used a different solution. The viewId in this case should differentiate svgLines in separate tabs. So $ is fine here. :)

Issue #2548, opened 2017-07-21
Cannot add non primitive types for struct fields


maheeka:
struct Person {
    string name;
    int age;
    Address person_address;
    string[] codes;
    json<Person> j;
}

viewed as : 
<img width="658" alt="image" src="https://user-images.githubusercontent.com/1029806/28462298-3f19b9fe-6e38-11e7-8899-5a7aab478257.png">

typing type fields gives an error as invalid type
<img width="1029" alt="image" src="https://user-images.githubusercontent.com/1029806/28462313-5076d524-6e38-11e7-9d93-cfd676b08393.png">

PR #2565, opened 2017-07-21
Fixing the issue when add non primitive types for struct fields

NatashaWso2:


NatashaWso2:
This fixes issue: https://github.com/ballerinalang/composer/issues/2548

PR #2607, opened 2017-07-24
load fonts for source view

pahans:


modules/web/index.html
```diff
@@ -28,6 +28,9 @@
 
     <script src="lib/css-browser-selector_0.4.0/css_browser_selector.js"></script>
 
+    <!-- loading monospaced font for source view -->
+    <link href="https://fonts.googleapis.com/css?family=Source+Code+Pro" rel="stylesheet">
```
hemikak: Can we download the font, keep it in lib and then refer to it. Reason is we cannot rely on external resources as the developing environment may not have internet or may have restricted network access.

hemikak: The font uses OFL(Source Code Pro can be adapted and redistributed according to the terms of the Open Font License (OFL) agreement.[1]) license. @hevayo should we be concerned ?

[1] - https://fonts.google.com/specimen/Source+Code+Pro

PR #2620, opened 2017-07-24
Fix issue No: 2605

### LINKED ISSUE #2605: Ordering issue in the client connector
Release - 0.9

**Steps**
1. Add a service
2. Add a HTTP client connector

**Issue**
The endpoint will be added at the bottom by default, so the reply statement is at top

endpoint

nadeeshaan:
Resolves #2605

### DIFF
--- a/modules/web/js/ballerina/components/panel-decorator.jsx
+++ b/modules/web/js/ballerina/components/panel-decorator.jsx
@@ -298,18 +298,36 @@ class PanelDecorator extends React.Component {
     }
 
     onDropZoneActivate(e) {
-        const dragDropManager = this.context.dragDropManager,
-            dropTarget = this.props.dropTarget,
-            dropSourceValidateCB = this.props.dropSourceValidateCB;
+        const dragDropManager = this.context.dragDropManager;
+        const dropTarget = this.props.dropTarget;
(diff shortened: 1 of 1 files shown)


modules/web/js/ballerina/components/panel-decorator.jsx
```diff
@@ -298,18 +298,36 @@ class PanelDecorator extends React.Component {
...
 
     onDropZoneActivate(e) {
-        const dragDropManager = this.context.dragDropManager,
-            dropTarget = this.props.dropTarget,
-            dropSourceValidateCB = this.props.dropSourceValidateCB;
+        const dragDropManager = this.context.dragDropManager;
+        const dropTarget = this.props.dropTarget;
+        const dropSourceValidateCB = this.props.dropSourceValidateCB;
+        const droppedNodeIndexCallBack = function () {
+            if (BallerinaASTFactory.isConnectorDeclaration(dragDropManager.getNodeBeingDragged())) {
```
hemikak: Cant we move the logic to ConnectorDeclaration ast implementation ?

nadeeshaan: Here the consideration was, that the panel is the one who should decide which should be the index. Connector declaration should not position himself because based on the drop zone it is being positioned. WDYT?

PR #2641, opened 2017-07-31
Stop showing empty page when there is a position/dimension calc error.

manuranga:
Error is logged and source view is shown.

### DIFF
--- a/modules/web/js/ballerina/components/diagram.jsx
+++ b/modules/web/js/ballerina/components/diagram.jsx
@@ -20,8 +20,6 @@ import React from 'react';
 import log from 'log';
 import PropTypes from 'prop-types';
 import CanvasDecorator from './canvas-decorator';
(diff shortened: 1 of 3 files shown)


modules/web/js/ballerina/components/diagram.jsx
```diff
@@ -58,65 +55,76 @@ class Diagram extends React.Component {
...
+
     /**
      * @override
      * @memberof Diagram
      */
     render() {
-        // Following is how we render the diagram.
-        // 1. We will visit the model tree and calculate width and height of all
-        //    the elements. We will use DimensionCalcVisitor.
-        this.props.model.accept(this.dimentionCalc);
```
hevayo: @manuranga as we discussed shall we keep the diagram calculations inside the diagram.

As an alternative inside the diagram render we can display an error message instead of rendering the diagram ?

hevayo:
Closing the PR. fix will be implemented to render an error message if the calculations fail.

PR #2674, opened 2017-08-09
Fix Issue 2298

### LINKED ISSUE #2298: "Error in BLang parser rest service for composer" when reply an element of a message type array
It works when the line 33 is removed
nullcomposer

nadeeshaan:
Resolves #2298


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/rest/datamodel/BLangExpressionModelBuilder.java
```diff
@@ -699,7 +699,20 @@ public void visit(FieldBasedVarRefExpr fieldBasedVarRefExpr) {
 
     @Override
     public void visit(IndexBasedVarRefExpr indexBasedVarRefExpr) {
-
+        StringBuffer buffer = new StringBuffer();
```
hemikak: We should use StringBuilder as StringBuffer is old.

PR #2685, opened 2017-08-10
Cached the builtin packages.

hemikak:
Builtin packages are loaded at first time only, then it is retrieved from cache.


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/LangServerManager.java
```diff
@@ -477,12 +482,14 @@ private void getBuiltInPackages(Message message) {
         if (message instanceof RequestMessage) {
             JsonObject response = new JsonObject();
             // Load all the packages associated the runtime
-            Map<String, ModelPackage> packages = Utils.getAllPackages();
-            LangServerManager.this.setPackages(packages.entrySet());
+            if (null == builtInNativePackages) {
```
maheeka: any specific gain of using null == over == null?

hemikak: No added advantage. Flipped

PR #2754, opened 2017-08-11
Allow empty struct input fields

pahans:


modules/web/js/ballerina/components/struct-definition-item.jsx
```diff
@@ -160,17 +159,28 @@ class StructDefinitionItem extends React.Component {
                         this.setState({
                             newDefaultValueEditing: false,
                         });
-                        model.setValue(this.state.newDefaultValue || value);
+                        if (!this.state.newDefaultValue) {
+                            const valueArrayId = model.getChildren() && model.getChildren()[1] && model.getChildren()[1].id;
```
maheeka: can we use model.getRightExpression() here?

Issue #2619, opened 2017-07-24
[Swagger] Multiple HTTP methods for a resource are not supported in swagger


dilinisg:
Release 0.91

If we implement a resource like below, only one HTTP method will be displayed in the swagger view
```
@http:POST {}
@http:GET {}
@http:Path {value:"/book/{country}"}
    
resource passthrough (){
}
```

hemikak:
Waiting till https://github.com/ballerinalang/ballerina/issues/3073 is resolved.

hemikak:
Ballerina does support multiple http method verbs. See "Support for multiple HTTP verbs in generated Swagger definitions." @ballerina-dev@googlegroups.com on how this is supported in the swagger editor.

PR #2777, opened 2017-08-16
Added swagger support for resources with multiple http methods.

hemikak:


hemikak:
Fixes https://github.com/ballerinalang/composer/issues/2619

PR #2785, opened 2017-08-18
Close web socket connection after all the language server tests

sumuditha-viraj:


modules/web/js/langserver/langserver-channel.js
```diff
@@ -31,6 +31,12 @@ class LangserverChannel extends EventChannel {
         if (_.isNil(args.endpoint)) {
             throw 'Invalid Endpoint';
         }
+        // assign provided ws close handler.
+        if (!_.isNil(args.clientController.options)){
+            if (!_.isNil(args.clientController.options.wsCloseEventHandler)) {
```
pahans: shall we use `if  (args.clientController.options && args.clientController.options.wsCloseEventHandler) {` here?

sumuditha-viraj: Done.

PR #2797, opened 2017-08-19
generate type cast and conversion expressions during mapping in trans…

maheeka:
…form stmt


maheeka:
Depends on https://github.com/ballerinalang/ballerina/pull/3230/

PR #2907, opened 2017-08-23
Bug fixes3

nadeeshaan:


muthulee:
Please leave a better PR name.

nadeeshaan:
@muthulee sorry for the bad practice. Will make sure from next time onwards

PR #2956, opened 2017-08-29
Remove gen directory

sumuditha-viraj:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/common/StringUtil.java
```diff
@@ -0,0 +1,66 @@
...
+     * <p>
+     * Note: This might be replaced by utility method from commons-lang or guava someday
+     * if one of those libraries is added as dependency.
+     * </p>
+     *
+     * @param array     The array of strings
+     * @param separator The separator
+     * @return the resulting string
+     */
+    public static String join(String[] array, String separator) {
```
maheeka: Can't we remove these utils ?

Issue #2971, opened 2017-08-31
[Transform] Invalid mappings are allowed


maheeka:
E.g. : Blob to boolean mapping should be disallowed.

PR #2972, opened 2017-08-31
Avoid invalid mappings based on the type lattice

maheeka:
Fixes : https://github.com/ballerinalang/composer/issues/2971

### DIFF
--- a/modules/web/js/ballerina/diagram/views/default/components/transform/transform-node-manager.js
+++ b/modules/web/js/ballerina/diagram/views/default/components/transform/transform-node-manager.js
@@ -48,10 +48,29 @@ class TransformNodeManager {
      */
     getCompatibility(source, target) {
         const compatibility = this._typeLattice.getCompatibility(source, target);
-        if (!compatibility || (source === target)) {
+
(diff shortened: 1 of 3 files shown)


modules/web/js/ballerina/diagram/views/default/components/transform/transform-node-manager.js
```diff
@@ -48,10 +48,24 @@ class TransformNodeManager {
...
+        // If the source and target are builtin types, we consider the mapping illegal
+        // if it is not available in the type lattice.
+        const builtInTypes = this._environment.getTypes();
+        if (!compatibility && builtInTypes.includes(source) && builtInTypes.includes(target)) {
+            return undefined;
+        }
+        return { safe: true, type: 'implicit' };
+    }
+
+    isConnectionValid(source, target) {
```
roonyh: This function is not used? I think because `!compatibility` works and looks good we don't this.

roonyh: ah. never mind. saw the usage.

Issue #2789, opened 2017-08-18
[Debugger] UI issues when removing arguments


dilinisg:
Release 0.92

**Issues**
1. No remove button for the first argument
2. A close button above the first arguments which is not working and also inconsistent

![args2](https://user-images.githubusercontent.com/1845370/29453841-b52af562-8428-11e7-851d-d133d49cb30a.png)

NipunaMarcus:
Issue is fixed hence closing the issue.

PR #2990, opened 2017-09-01
Remove app argument config window to have add argument button

NipunaMarcus:
Fixed issue: https://github.com/ballerinalang/composer/issues/2789

### DIFF
--- a/modules/web/js/debugger/tools.js
+++ b/modules/web/js/debugger/tools.js
@@ -177,6 +177,7 @@ class Tools extends EventChannel {
         this.container.on('click', '.btn-config', (e) => {
             e.preventDefault();
             e.stopPropagation();
+            this.appArgsDialog.find('#form-run-application-with-args .removable').remove();
             this.appArgsDialog.modal('show');
         });
         const self = this;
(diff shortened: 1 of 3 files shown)


modules/web/js/debugger/tools.js
```diff
@@ -212,16 +213,12 @@ class Tools extends EventChannel {
...
                 const uniqueId = this.getFileUniqueId(activeTab.getFile());
                 const argsString = this.application.browserStorage.get(`launcher-app-configs-${uniqueId}`) || '';
-                const currentArgs = argsString.split(' ');
+                const currentArgs = argsString ? argsString.split(' ') : [];
                 _.each(currentArgs, (arg, i) => {
-                    if (i === 0) {
-                        $('#form-run-application-with-args input[type=\'text\']').get(0).value = arg;
-                    } else {
-                        $(wrapper).append(`<div class="removable">
+                    $(wrapper).append(`<div class="removable">
```
pahans: seems there is a formatting issue here.

NipunaMarcus: No @pahans this is correcly formatted ... this code snippet was inside a else statement not i have remove that else so it is like this.

PR #2993, opened 2017-09-02
Dynamic Connector Icon Support

nadeeshaan:


modules/web/js/ballerina/diagram/views/default/components/image-util.jsx
```diff
@@ -63,6 +65,39 @@ class ImageUtils {
...
+    static getConnectorIcon(connectorName) {
+        const requestJson = {
+            connectorName: connectorName,
+            iconPath: 'testPath',
+        };
+        let data = {};
+        $.ajax({
+            type: 'POST',
+            context: this,
+            url: 'http://localhost:8289/file/connector/icon',
```
manuranga: can we use '/file/connector/icon' here so any future port changes will not brake this?

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/fileserver/FileContentProvider.java
```diff
@@ -0,0 +1,119 @@
...
+    @Produces(MediaType.APPLICATION_JSON)
+    public Response getConnectorIcon(ConnectorIconRequest connectorIconRequest, @Context Request request) {
+        JsonObject responseJson;
+
+        if (this.contextRoot != null && this.contextRoot.endsWith("/resources/composer/web")) {
+            logger.debug(connectorIconRequest.getIconPath());
+            logger.debug(connectorIconRequest.getConnectorName());
+
+            try {
+                String filePath = this.contextRoot + File.separator + "images" + File.separator + "connectors" +
```
manuranga: This may result in a https://www.owasp.org/index.php/Path_Traversal . ether test that it's not vulnerable or use `Path.startsWith` check to avoid.

nadeeshaan: This won't apply in this scenario. If you look at the full implementation, context root has been check,  as well as, the only parameter outside, is the connector icon request which is also sent from the composer client without any outside interference.

PR #3011, opened 2017-09-06
Added action menu for panels.

hemikak:
Added action menu for:
ServiceDefs,
ResourceDefs,
FunctionDefs,
ConnectorDefs,
ConnectionActionDefs,
AnnotationDefs,
StructDefs.

### DIFF
--- a/modules/web/js/ballerina/configs/app-context.js
+++ b/modules/web/js/ballerina/configs/app-context.js
@@ -16,14 +16,16 @@
  * under the License.
  */
 
-let _wrapper;
+import _ from 'lodash';
+
+let canvasOverlay;
 
 export function setCanvasOverlay(wrapper) {
-    if (wrapper !== undefined && wrapper !== null) {
-        _wrapper = wrapper;
+    if (!_.isNil(wrapper)) {
+        canvasOverlay = wrapper;
     }
 }
 
 export function getCanvasOverlay() {
-    return _wrapper;
+    return canvasOverlay;
 }
--- a/modules/web/js/ballerina/configs/app-context.js
+++ b/modules/web/js/ballerina/configs/app-context.js
@@ -19,7 +19,9 @@
 let _wrapper;
 
 export function setCanvasOverlay(wrapper) {
-    _wrapper = wrapper;
+    if (wrapper !== undefined && wrapper !== null) {
+        _wrapper = wrapper;
+    }
 }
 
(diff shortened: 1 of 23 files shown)


modules/web/js/ballerina/configs/app-context.js
```diff
@@ -19,7 +19,9 @@
 let _wrapper;
 
 export function setCanvasOverlay(wrapper) {
-    _wrapper = wrapper;
+    if (wrapper !== undefined && wrapper !== null) {
```
manuranga: may use `_.isNil`

modules/web/js/ballerina/diagram/views/default/components/annotation-definition.jsx
```diff
@@ -17,26 +17,211 @@
...
+    }
+
+    /**
+     * Gets the suggestions for attachment points for the definition.
+     * @returns {string[]} The suggestions.
+     * @memberof AnnotationDefinition
+     */
+    getAnnotationAttachmentPointsForSuggestions() {
+        const { environment } = this.context;
+        const suggestions = [];
```
manuranga: `const attachmentTypes = environment.getAnnotationAttachmentTypes();` will look much nicer since it's used twice.

modules/web/js/ballerina/diagram/views/default/components/resource-definition.jsx
```diff
@@ -44,20 +89,84 @@ class ResourceDefinition extends React.Component {
...
+    /**
+     * Creates the action menu.
+     * @memberof ResourceDefinition
+     */
+    createActionMenu() {
+        const model = this.props.model;
+        const designer = getDesigner(['default']);
+        const canvasOverlay = getCanvasOverlay();
+        // Recreating content
+        this.actionMenuWrapper = document.createElement('div');
```
manuranga: why are divs manually created instead of React creating them?

hemikak: canvasOverlay is an html node. There can be multiple actionMenuWrappers added to canvasOverlay added from multiple components(jsx) as each action menu can contain different menu items. If using react, only one actionMenuWrapper can be added to canvasOverlay using ReactDom. Cannot add multiple actionMenuWrappers using ReactDom to a canvasOverlay.

PR #3044, opened 2017-09-11
Serialize model to JSON.

manuranga:
Temp workaround by creating a temp file form stream. Also need following local change to run-time.

```
--- modules/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/Compiler.java	(date 1505129695000)
+++ modules/ballerina-lang/src/main/java/org/wso2/ballerinalang/compiler/Compiler.java	(revision )
@@ -76,7 +76,7 @@
         }
     }
 
-    private BLangPackage define(String sourcePkg) {
+    public BLangPackage define(String sourcePkg) {
         return pkgLoader.loadEntryPackage(sourcePkg);
     }
 ```

### DIFF
--- a/modules/web/js/ballerina/model/source-gen.js
+++ b/modules/web/js/ballerina/model/source-gen.js
@@ -372,8 +372,8 @@ export default function getSourceOf(node, pretty = false, l = 0) {
         case 'Invocation':
             if (node.expression && node.name.value && node.argumentExpressions) {
                 return getSourceOf(node.expression, pretty, l) + w() + '.' + w()
-                    + node.name.value + w() + '('
-                    + join(node.argumentExpressions, pretty, l, w, '', ',') + w() + ')';
+                 + node.name.value + w() + '('
+                 + join(node.argumentExpressions, pretty, l, w, '', ',') + w() + ')';
             } else if (node.packageAlias.value && node.name.value
                          && node.argumentExpressions) {
                 return w() + node.packageAlias.value + w() + ':' + w() + node.name.value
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: modules/services/workspace-service/pom.xml [hunk 1]
```diff
@@ -161,6 +165,10 @@
             <groupId>com.beust</groupId>
             <artifactId>jcommander</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.ballerinalang</groupId>
+            <artifactId>ballerina-lang</artifactId>
```
hemikak: Dependency added twice ?

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/rest/datamodel/BLangFileRestService.java
```diff
@@ -145,13 +171,125 @@ public Response optionsParseFragment() {
      * @throws IOException
      */
     private static String parseJsonDataModel(InputStream stream, java.nio.file.Path filePath) throws IOException {
+        String tmpDir = System.getProperty("java.io.tmpdir");
+        CompilerContext context = new CompilerContext();
+        CompilerOptions options = CompilerOptions.getInstance(context);
+        File tmpFile = File.createTempFile("composer.tmp.", ".bal", new File(tmpDir));
+        OutputStream outputStream = new FileOutputStream(tmpFile);
+        IOUtils.copy(stream, outputStream);
+        outputStream.close();
```
hemikak: You can use https://commons.apache.org/proper/commons-io/javadocs/api-2.5/org/apache/commons/io/IOUtils.html#closeQuietly(java.io.Closeable...) here so that it wont throw the IOException.

Only if it is needed.

PR #3089, opened 2017-09-19
Language server refactor

nadeeshaan:
This includes the initial refactoring changes for the language server

### DIFF
--- a/modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/LangServerManager.java
+++ b/modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/LangServerManager.java
@@ -25,6 +25,7 @@ import com.google.gson.JsonParser;
 import com.google.gson.internal.LinkedTreeMap;
 import io.netty.channel.Channel;
 import io.netty.handler.codec.http.websocketx.TextWebSocketFrame;
+import org.antlr.v4.runtime.DefaultErrorStrategy;
 import org.ballerinalang.BLangASTBuilder;
 import org.ballerinalang.composer.service.workspace.Constants;
 import org.ballerinalang.composer.service.workspace.common.Utils;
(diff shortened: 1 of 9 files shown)


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/LangServerManager.java
```diff
@@ -511,42 +525,42 @@ private void getBuiltInPackages(Message message) {
...
+            String textContent = posParams.getText();
 
+            SuggestionsFilterDataModel filterDataModel = new SuggestionsFilterDataModel();
+            CapturePossibleTokenStrategy errStrategy = new CapturePossibleTokenStrategy(position, filterDataModel);
+
+            CompilerContext compilerContext = new CompilerContext();
+            compilerContext.put(DefaultErrorStrategy.class, errStrategy);
+
+            options = CompilerOptions.getInstance(compilerContext);
+            options.put(SOURCE_ROOT, "/home/nadeeshaan/Desktop");
```
sumuditha-viraj: Can we avoid hardcoded Path?

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/LangServerManager.java
```diff
@@ -511,42 +525,42 @@ private void getBuiltInPackages(Message message) {
...
+            options.put(COMPILER_PHASE, "typeCheck");
+
+            Compiler compiler = Compiler.getInstance(compilerContext);
+            // here we need to compile the whole package
+            BLangPackage bLangPackage = compiler.compile("test.bal");
```
sumuditha-viraj: Can we avoid hardcoded file names?

nadeeshaan: This will be replaced accordingly, with the client side request path is completed and this hardcoded path is temporarily added.

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/LangServerManager.java
nadeeshaan: This will be replaced accordingly, with the client side request path is completed and this hard-coded path is temporarily added.

PR #3101, opened 2017-09-21
Implementation of property views for services and connectors and making services a top level construct 

NatashaWso2:


modules/web/js/ballerina/diagram/views/default/components/connector-properties.jsx
```diff
@@ -0,0 +1,252 @@
...
+        }
+        return value;
+    }
+
+    /**
+     * Get full package path of connector
+     * @param pkgName
+     * @returns {string}
+     */
+    getFullPkgPath(pkgName) {
```
hevayo: We should get this dynamically.

NatashaWso2: This was hard coded because we cannot get the full package path of the connector from the composer design view atm. This is an issue with the fragment parser and can be resolved with the implementation of the new semantic analyser.

PR #3146, opened 2017-09-23
Initial commit for rendering the package declaration, global definitions and structs

NatashaWso2:


modules/web/js/ballerina/diagram2/views/default/components/nodes/image-util.jsx
```diff
@@ -0,0 +1,68 @@
+/**
```
hevayo: Lets use the one inside diagram2 directory

PR #3148, opened 2017-09-23
Changed order of imports so we can override css rules from libraries we add

roonyh:
Fixed more glitches in transform expanded view

### DIFF
--- a/modules/web/src/index.js
+++ b/modules/web/src/index.js
@@ -1,10 +1,3 @@
-
-import log from 'log';
-import _ from 'lodash';
-import { fetchConfigs } from 'api-client/api-client';
-import Application from './core/app';
-import defaultConfig from './config';
-
 // importing for side effects only
 import 'bootstrap';
 import 'theme_wso2';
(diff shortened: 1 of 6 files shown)


roonyh:
Before: 
screenshot from 2017-09-23 13-00-51

Now:
screenshot from 2017-09-23 14-34-24

modules/web/src/index.js
```diff
@@ -1,17 +1,15 @@
...
 import 'theme_wso2';
 import 'bootstrap/dist/css/bootstrap.min.css';
 import 'bootstrap/dist/css/bootstrap-theme.min.css';
 import 'font-ballerina/css/font-ballerina.css';
 
+import log from 'log';
+import _ from 'lodash';
+import { fetchConfigs } from 'api-client/api-client';
+import Application from './core/app';
+import defaultConfig from './config';
```
roonyh: @dakshika @kaviththiranga I Changed order of imports so we can override css rules from libraries we add. This is fine ne?

PR #3199, opened 2017-09-28
Add debugger tools to main toolbar

pahans:


modules/web/src/core/toolbar/components/Tool.jsx
```diff
@@ -12,13 +12,17 @@ class Tool extends React.Component {
      * @inheritdoc
      */
     render() {
-        const { tool: { icon, commandID, commandArgs = {}, isActive = () => true },
+        const { tool: { icon, commandID, commandArgs = {}, isActive = () => true, isVisible = () => true },
                 dispatch } = this.props;
+
+        if (!isVisible()) {
+            return (<noscript />);
```
dakshika: Is there any unsupported things here?

pahans: `<noscript />` is used when we dont want to render anything in react.

pahans: more info https://stackoverflow.com/questions/31231913/does-react-have-a-null-element/31232642

pahans: seems they have updated API to use null instead. https://facebook.github.io/react/blog/2014/07/17/react-v0.11.html#rendering-to-null

PR #3219, opened 2017-09-29
Fixed parsing empty content in backend.

hemikak:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/rest/datamodel/BLangFileRestService.java
```diff
@@ -167,39 +168,43 @@ private static String parseJsonDataModel(BFile bFile) throws IOException, Invoca
...
         JsonElement modelElement = generateJSON(compilationUnit);
         JsonObject result = new JsonObject();
         result.add("model", modelElement);
         result.add("diagnostics", diagnosticsJson);
-        result.add("packageInfo", packageInfo);
-
+    
+        // Add 'packageInfo' only if there are any packages.
+        Optional<ModelPackage> packageInfoJson = modelPackage.values().stream().findFirst();
+        if (packageInfoJson.isPresent()) {
```
hemikak: @sumuditha-viraj , packageInfo will be passed only if it is available. I check the JS code, we have a null check there[1]. So I think it will be fine to send if packageInfo is available.

[1] - https://github.com/ballerinalang/composer/blob/semantic-refactor/modules/web/js/ballerina/views/ballerina-file-editor.jsx#L405

sumuditha-viraj: +1

PR #3237, opened 2017-10-01
Make transform operators drag drop-able

roonyh:


modules/web/js/ballerina/ast/default-ast-factory.js
```diff
@@ -274,10 +275,14 @@ DefaultASTFactory.createAggregatedAssignmentStatement = function (args) {
     return ASTFactory.createAssignmentStatement(args);
 };
 
+let outPutVarCounter = 0;
 DefaultASTFactory.createTransformAssignmentOperatorStatement = function (args = {}) {
```
maheeka: Let's move this to TransformFactory.js

PR #3264, opened 2017-10-03
Rendering services, resources and the property window 

NatashaWso2:


modules/web/js/ballerina/diagram2/views/default/components/decorators/canvas-decorator.jsx
```diff
@@ -55,6 +56,18 @@ class CanvasDecorator extends React.Component {
                     {/* This space is used to render html elements over svg*/ }
                 </div>
                 {(this.props.annotations && this.props.annotations.length > 0) ? this.props.annotations : null }
+                <CSSTransitionGroup
```
hevayo: we should apply transitions for individual components. lets refactor after merge.

modules/web/js/ballerina/diagram2/views/default/components/nodes/resource-transport-link.jsx
```diff
@@ -0,0 +1,46 @@
+/**
```
hevayo: we should move this file in to decorators

modules/web/js/ballerina/diagram2/views/default/components/nodes/service-definition.css
```diff
@@ -0,0 +1,33 @@
+.add-resource-button {
```
hevayo: Lets rename the css file to match the node component name.

PR #3280, opened 2017-10-04
Use fragment parser to create dragging node

kaviththiranga:


modules/web/src/core/editor/views/EditorTabs.jsx
```diff
@@ -217,15 +217,15 @@ class EditorTabs extends View {
...
-                            {title}
                             <button
                                 type="button"
                                 className="close close-tab pull-right"
                                 onClick={() => this.onTabClose(editor)}
                             >
                                 ×
                             </button>
+                            <i className={`fw fw-${icon} tab-icon`} />
+                            {title}
```
hemikak: Was this change because the close button came before the title ? If so we need to fix it using float:left I think.

hevayo: This is fine I since we use the pull-right for close button.

PR #3283, opened 2017-10-04
updated tree nodes

hevayo:


modules/web/js/ballerina/diagram2/diagram.jsx
```diff
@@ -65,6 +65,7 @@ class Diagram extends React.Component {
      * @memberof Diagram
      */
     render() {
+        console.log(this.props.model);
```
hemikak: Need to remove the console log.

PR #3317, opened 2017-10-06
fix remove transform structs and variables

keizer619:


modules/web/js/ballerina/diagram2/views/default/components/transform/transform-node-mapper.js
```diff
@@ -758,7 +758,9 @@ class TransformNodeMapper {
      * @memberof TransformNodeMapper
      */
     removeSourceType(type) {
-        this._transformStmt.removeInput(type);
+        _.remove(this._transformStmt.inputs, (input) => {
```
maheeka: removeInputs(type) should work

modules/web/js/ballerina/diagram2/views/default/components/transform/transform-node-mapper.js
```diff
@@ -876,6 +882,9 @@ class TransformNodeMapper {
                     const simpleVarRefExpression = TransformFactory.createVariableRefExpression(outputVarName);
                     stmt.replaceVariablesByIndex(index, simpleVarRefExpression, true);
                 }
+            } else if (TreeUtil.isFieldBasedAccessExpr(stmt.getVariables()[0])
+                        && stmt.getVariables()[0].getExpression().getVariableName().getValue() === expStr) {
```
maheeka: getSource() ??

keizer619: @maheeka 
when removing a struct, expr will be struct object name. if we use getSource() it will return <structName>.<fieldName>. So instead this is checking VariableName against expr for FieldBasedAccessExpr

PR #3320, opened 2017-10-06
Changes for latest core updates

nadeeshaan:


hevayo:
implementation has differed

PR #3334, opened 2017-10-07
Dynamic endpoint update for action invocation

nadeeshaan:


modules/web/js/ballerina/diagram2/views/default/components/decorators/arrow-decorator.jsx
```diff
@@ -41,8 +41,7 @@ class Arrow extends React.Component {
         const deltaY = end.y - start.y;
         const rad = Math.atan2(deltaY, deltaX);
         const deg = rad * (180 / Math.PI);
-
-        return deg;
+        return deg - this.props.backward ? 180 : 0;
```
hevayo: can we put 180 in to congigs

PR #3335, opened 2017-10-07
Transform Improvements

maheeka:


modules/web/js/ballerina/env/environment.js
```diff
@@ -179,6 +182,25 @@ class BallerinaEnvironment extends EventChannel {
     }
 
     /**
+     * Initialize operator lattice
+     */
+    initializeOperatorLattice() {
+        return new Promise((resolve, reject) => {
```
roonyh: No need to create a new promise here. You can return the promise received from `getOperatorLatice`.
~~~
return getOperatorLattice()
                 .then((operatorLatticeJson) => {
                     if (operatorLatticeJson) {
                         this._operatorLattice.initFromJson(operatorLatticeJson);
                     } else {
                         log.error('Error while fetching operator lattice');
                     }
                 })
~~~

modules/web/js/type-lattice/operator-lattice.js
```diff
@@ -0,0 +1,97 @@
...
+        const binaryLattice = {};
+
+        const unaryLatticeJson = operatorLatticeJson.filter(edge => edge.type === 'unary');
+        const binaryLatticeJson = operatorLatticeJson.filter(edge => edge.type === 'binary');
+
+        unaryLatticeJson.forEach((edge) => {
+            if (!edge.visited) {
+                const operator = edge.operator;
+                unaryLattice[operator] = {};
+                if (!edge.visited) {
```
roonyh: The check `if (!edge.visited)` is already done in line 47. No need to do it again right?  
Also we can do an early return,
`if (edge.visited) { return; }`

modules/web/js/type-lattice/operator-lattice.js
```diff
@@ -0,0 +1,97 @@
...
+                    });
+                }
+            }
+        });
+
+        binaryLatticeJson.forEach((edge) => {
+            if (!edge.visited) {
+                const operator = edge.operator;
+                binaryLattice[operator] = {};
+                if (!edge.visited) {
```
roonyh: Same as above, no need to check again?

modules/web/js/ballerina/env/environment.js
maheeka: There are no promises returning from getOperatorLattice as we have hardcoded it for the moment. Shall we leave this for now? I'll add a TODO

modules/web/js/type-lattice/operator-lattice.js
maheeka: This is a two level check. We are iterating over the same list twice and hence the two checks.

modules/web/js/type-lattice/operator-lattice.js
maheeka: Same as above

modules/web/js/type-lattice/operator-lattice.js
roonyh: um. but then the second if check should be inside the second forEach ne? These two ifs are in the same iteration. :thinking:

modules/web/js/ballerina/env/environment.js
roonyh: `Promise.resolve` returns a promise here https://github.com/ballerinalang/composer/pull/3335/files#diff-e30e1d6640d67749adfae49b49d69af0R293

PR #3339, opened 2017-10-08
Add the relevant import when an assignment statement with a function invocation is dragged in

roonyh:


modules/web/js/ballerina/model/tree-builder.js
```diff
@@ -58,6 +58,10 @@ class TreeBuilder {
         // 4. If there is a more specific node defined based on node kind we will use
         //    that object instead of generic object.
 
+        if (!treeNodes) {
+             treeNodes = requireAll(require.context('./tree/', true, /\.js$/));
+        }
+
```
roonyh: Had to take this in to the build function, because otherwise `tree-builder` can't be imported inside `compilation-unit-node`. The cyclic import results in `CompilationUnitNode` not being added to `treeNodes`

modules/web/js/ballerina/model/tree/compilation-unit-node.js
```diff
@@ -20,10 +20,32 @@ import _ from 'lodash';
...
+            }
+
+            if (TreeUtils.isAssignment(e.data.node) && TreeUtils.isInvocation(e.data.node.getExpression())) {
+                const fullPackageName = e.data.node.getExpression().getFullPackageName();
+
+                const value = 'import ' + fullPackageName + ';\n';
+                parseContent(value)
+                    .then((jsonTree) => {
+                        if (jsonTree.model.topLevelNodes[0]) {
+                            this.addImport(TreeBuilder.build(jsonTree.model.topLevelNodes[0]));
```
manuranga: I don't think its a good idea for the Tree to build itself, this leads to circular dependency like we had earlier. why cant we keep the builder separate.

modules/web/js/ballerina/model/tree/compilation-unit-node.js
```diff
@@ -20,10 +20,32 @@ import _ from 'lodash';
...
+        super();
+        this.on('tree-modified', (e) => {
+            if (e.type !== 'child-added') {
+                return;
+            }
+
+            if (TreeUtils.isAssignment(e.data.node) && TreeUtils.isInvocation(e.data.node.getExpression())) {
+                const fullPackageName = e.data.node.getExpression().getFullPackageName();
+
+                const value = 'import ' + fullPackageName + ';\n';
```
maheeka: Should there be a current package check before adding the import?

roonyh: ah. yeah. will add it.

modules/web/js/ballerina/views/ballerina-file-editor.jsx
```diff
@@ -181,6 +187,29 @@ class BallerinaFileEditor extends React.Component {
     }
 
     /**
+     * Adds relevent imports needed to be automatically imported When a node (eg: a function invocation) is dragged in
+     * @param {Node} node the node added
+     */
+    addAutoImports(node) {
+        if (!TreeUtils.isAssignment(node) || !TreeUtils.isInvocation(node.getExpression())) {
```
maheeka: we need to check for var def statements also here

roonyh: Um. Does dragged in invocations create var def statements that require new imports to be added?  Aren't they always assignments?

maheeka: General tool palette has var def statements. But transform is assignments. Since we are making it generic, should add both.

roonyh: Got it. Updated PR

PR #3694, opened 2017-10-13
Open left panel when running

pahans:


modules/web/src/core/editor/constants.js
```diff
@@ -4,6 +4,7 @@ export const COMMANDS = {
     UNDO: 'undo',
     REDO: 'redo',
     ACTIVATE_EDITOR_FOR_FILE: 'activate-editor-for-file',
+    SHOW_LEFT_PANEL: 'show-left-panel',
```
kaviththiranga: this is not needed right, it's defined in layout events (layout-plugin/constant.js)

PR #3873, opened 2017-10-17
Fix sync line numbers

pahans:


pahans:
@NatashaWso2 can you verify this fix with your changes here https://github.com/ballerinalang/composer/commit/470f5b8ba68a11d02f6ad9bdfe2d1adcfa071813#diff-057585a56c95f7825d7ca46fc0c973daR108

modules/web/js/ballerina/model/tree/node.js
```diff
@@ -92,21 +92,20 @@ class Node extends EventChannel {
         // eslint-disable-next-line guard-for-in
         for (const childName in this) {
             if (childName !== 'parent' && childName !== 'position' && childName !== 'ws') {
-                const child = this[childName];
-                const child2 = newTree[childName];
-                if (!child2) {
+                if (!this[childName] || !newTree || !newTree[childName]) {
                     continue;
```
hevayo: This is an invalid case shall we put a error log ?

Issue #3651, opened 2017-10-12
File not added through directory structure when typed in without .bal extension


erandidesilva:
Browser: Chrome Version 61.0.3163.100 (Official Build) (64-bit)

**Steps**
1. Go to diretory listing
2. Open a program directory
3. Create new file by right click -> New file
4. Type in a name without .bal extension

The file is not created. 

**Expected result**
The file should be created with the .bal extension added, or an option should be given to user to select the file type

kaviththiranga:
Fixed with https://github.com/ballerinalang/composer/pull/3892

PR #3892, opened 2017-10-17
Implicitly add bal extension while adding files from tree

kaviththiranga:
TODO: We need to improve plugin arch to contribute supported extension
types for new file/save file wizards similar to editor contribs.

Fixes
https://github.com/ballerinalang/composer/issues/3651


hemikak:
@Kavith, will this only affect file adding.

I have a dialog which opens jsons and yamls. - https://github.com/ballerinalang/composer/blob/master/modules/web/src/plugins/import-swagger/plugin.js#L186

kaviththiranga:
@hemikak, yeah  this won't have an effect on wizards. This is only for tree nodes in explorer.

Issue #3668, opened 2017-10-12
Although file is renamed, it doesn't take effect


erandidesilva:
Browser: Chrome Version 61.0.3163.100 (Official Build) (64-bit)

**Steps**
1. Open directory
2. Open/ create a file
3. Open the file
4. Rename the file 

Although file is renamed, the open file name doesn't change. When double clicked on the renamed file, another file opens in another tab.

kaviththiranga:
file is renamed properly and effects are reflected in file system too. Issues is when the renamed file is opened in workspace. We need to add logic to intercept rename/delete actions on opened-files in workspace and make sure the tab reflects the changes. ATM after rename, in old tab for old file, if a change is done and saved a new file with old name will be created at same location.

kaviththiranga:
As an initial solution, we can forcefully close exiting tab and open a new tab upon renaming while blocking user from rename (also delete) if tab is in dirty state.

kaviththiranga:
@hevayo WDYT?

kaviththiranga:
Fixed with https://github.com/ballerinalang/composer/pull/3919

PR #3919, opened 2017-10-18
Handle rename/delete from tree when file is already opened in editor

kaviththiranga:


kaviththiranga:
Fixes https://github.com/ballerinalang/composer/issues/3668

modules/web/src/core/editor/dialogs/OpenedFileDeleteConfirmDialog.jsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
```
maheeka: License header

kaviththiranga: Fixed

Issue #3713, opened 2017-10-13
Try-it cursor icon changes to hand click format when placed on response parameters


erandidesilva:
Broswer: Chrome version 61.0.3163.100 (Official Build) (64-bit)

Please see the attached image. The cursoir icon should not be the hand rather the pointer, since there are no linkables to follow.

![com_0 94_tryit_3](https://user-images.githubusercontent.com/15624590/31534804-208445e8-b016-11e7-8f3a-ac25e84f1248.png)

maheeka:
Can you update the screenshot please? Unable to see the said hand pointer in the attached screenshot.

erandidesilva:
Seemingly I can't get it captured to a screenshot. But please find the below steps

1. Go to Run
2. Go to Try-it
3. Hover your mouse on "Time Consumed" label's  value.

Then you will be able to see the cursor icon changing in to hand.  On Reponse Code, Request URL, Request Headers, Response headers the same behaviour can be observed.

PR #3948, opened 2017-10-18
Applying the CSS styles of the prop window elements using a parent selector

NatashaWso2:


hemikak:
Fixes https://github.com/ballerinalang/composer/issues/3713

PR #3971, opened 2017-10-19
Improving unit test by comparing jsons as objects.

manuranga:


modules/services/workspace-service/src/test/java/org/ballerinalang/composer/service/workspace/rest/datamodel/BLangFragmentParserTest.java
```diff
@@ -68,7 +70,7 @@ public void testFragmentParse(String type, String caseName, String sourcePath, S
         fragment.setExpectedNodeType(type);
 
         String result = BLangFragmentParser.parseFragment(fragment);
-        Assert.assertEquals(result, readFromFile(expectedPath),
+        Assert.assertEquals(jsonParse.parse(result), jsonParse.parse(readFromFile(expectedPath)),
```
maheeka: Should we do this for other tests as well?

Issue #3661, opened 2017-10-12
Swagger is not generated when there are multiple resources


yasassri:
![screenshot from 2017-10-12 20-03-03](https://user-images.githubusercontent.com/7681361/31501678-8686523e-af88-11e7-9e7a-d78781e00b02.png)

nadeeshaan:
Check with the latest build and could not reproduce the issue. (Refer the below snapshot)
Hence closing the issue. If there is a specific source please reopen the issue

![selection_116](https://user-images.githubusercontent.com/1329674/31613266-3477c3ea-b2a0-11e7-92cd-1e2802496dc8.png)

yasassri:
Hi @nadeeshaan,

I encountered this issue again, couldn't  see any error logs though. 

![screenshot from 2017-10-17 05-31-04](https://user-images.githubusercontent.com/7681361/31640495-758fc2a4-b2fc-11e7-8702-dd3119249241.png)

hemikak:
@yasassri , can you attach the bal source for this scenario ?

hemikak:
@yasassri  Thanks

PR #3978, opened 2017-10-19
Removed @http:PathParam and @http:QueryParam support.

hemikak:
Fixes https://github.com/ballerinalang/composer/issues/3661


hemikak:
Fixes https://github.com/ballerinalang/composer/issues/4018  - to test more.

PR #4067, opened 2017-10-20
Fix about-dialog close btn not clickable & patch version in about dialog version

kaviththiranga:


modules/web/src/plugins/help/dialogs/about-dialog.jsx
```diff
@@ -36,7 +36,19 @@ class AboutDialog extends React.Component {
...
+                    <div>
+                        <div className="brand">
+                        <img
+                            src="images/BallerinaLogo.svg"
+                            alt="Ballerina Composer"
+                            className="logo"
+                        />
+                        <span className="appname">Composer</span>
+                        </div>
+                        <div className="version">v0.93.SNAPSHOT</div>
```
pahans: this should be 0.93.1 right?

PR #4139, opened 2017-10-23
Add semantic error to model

sumuditha-viraj:


modules/web/js/ballerina/diagram2/diagram.jsx
```diff
@@ -112,6 +114,10 @@ class Diagram extends React.Component {
             .setWorkerInvocationPositionSyncUtil(getInvocationArrowPositionUtil(this.props.mode));
         this.props.model.accept(this.invocationArrowPositionVisitor);
 
+        // 7. Now we will enrich the model with Semantic errors.
```
hevayo: Shall we do the enrichment after the validation call I think it is a responsibility of the ballerina file editor.

sumuditha-viraj: Yeah. That make sense. But the only place these errors are being used is the "diagram" react component. Reason is, source editor has its own way of error showing mechanism. Source editor will not be using errors from the model. So having this error enrichment in the "diagram" might separate the concerns. I'll move this error enrichment to Ballerina file editor so that we will be able to use this enriched model in some other editor views, if we came across such an use case in future.

pahans:
Avoid merge commits by using rebase :) https://randyfay.com/content/simpler-rebasing-avoiding-unintentional-merge-commits

PR #4199, opened 2017-10-26
Added validation to recognize main function

### LINKED ISSUE #1927: Returns and Params in main function should be uneditable
image

Main function has a fixed signature as : 

`function main(string[] args)`

Therefore, params and returns should not be modifiable (similar to function name)

NipunaMarcus:
Fix #1927


modules/web/js/ballerina/diagram2/views/default/components/decorators/panel-decorator.jsx
```diff
@@ -81,7 +81,14 @@ class PanelDecorator extends React.Component {
     }
 
     onTitleClick() {
-        if (this.props.model.getKind() === 'Function' && this.props.model.getName().value === 'main') {
+        const model = this.props.model;
+        if (model.getKind() === 'Function'
+            && model.getName().value === 'main'
```
hevayo: Can we move this logic to tree util and use them in other places. ie isMainFunction ...

PR #4200, opened 2017-10-26
Adding bound function resolving for language server

### LINKED ISSUE #4126: No content assist in http services
screenshot from 2017-10-22 11-55-31

nadeeshaan:
With the added functionality, bound functions can be resolved as follows.

selection_124

Resolves #4126

### DIFF
--- a/modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/util/completion/filters/PackageActionAndFunctionFilter.java
+++ b/modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/util/completion/filters/PackageActionAndFunctionFilter.java
@@ -44,10 +44,10 @@ public class PackageActionAndFunctionFilter implements SymbolFilter {
         String delimiter = tokenStream.get(delimiterIndex).getText();
         ArrayList<SymbolInfo> returnSymbolsInfoList = new ArrayList<>();
 
-        if (delimiter.equals(".")) {
+        if (".".equals(delimiter)) {
             // If the delimiter is "." then we are filtering the bound functions for the structs
             returnSymbolsInfoList.addAll(this.getBoundActionAndFunctions(dataModel, symbols, delimiterIndex));
-        } else if (delimiter.equals(":")) {
+        } else if (":".equals(delimiter)) {
             // We are filtering the package functions
             returnSymbolsInfoList.addAll(this.getActionsAndFunctions(dataModel, symbols, delimiterIndex));
         }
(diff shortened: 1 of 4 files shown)


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/util/completion/filters/PackageActionAndFunctionFilter.java
```diff
@@ -41,55 +41,83 @@
...
-            int initTokenIndex = this.getIndexOfTokenString(searchTokenString, 0, dataModel);
-            int equalTokenIndex = this.getIndexOfTokenString("create", initTokenIndex, dataModel);
-            colonTokenIndex = this.getIndexOfTokenString(":", equalTokenIndex, dataModel);
-        }
-
         ArrayList<SymbolInfo> returnSymbolsInfoList = new ArrayList<>();
 
-        if (colonTokenIndex > -1) {
-            String packageName = tokenStream.get(colonTokenIndex - 1).getText();
+        if (delimiter.equals(".")) {
```
sumuditha-viraj: I would always use - if (".".equals(delimiter)) instead of - if (delimiter.equals(".")). 
It will avoid NPEs. (Eg: delimiter == null scenario)

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/util/completion/filters/PackageActionAndFunctionFilter.java
```diff
@@ -41,55 +41,83 @@
...
+            // If the delimiter is "." then we are filtering the bound functions for the structs
+            returnSymbolsInfoList.addAll(this.getBoundActionAndFunctions(dataModel, symbols, delimiterIndex));
+        } else if (delimiter.equals(":")) {
```
sumuditha-viraj: Please see above comment.

nadeeshaan: fixed

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/util/completion/filters/PackageActionAndFunctionFilter.java
nadeeshaan: fixed

PR #4208, opened 2017-10-27
[Transform] Operand removal and test improvements

maheeka:
…e's default value

### DIFF
--- a/modules/web/js/ballerina/model/transform-factory.js
+++ b/modules/web/js/ballerina/model/transform-factory.js
@@ -16,6 +16,7 @@
  * under the License.
  */
 
+import log from 'log';
 import FragmentUtils from '../utils/fragment-utils';
 import TreeBuilder from './tree-builder';
 import TreeUtil from './tree-util';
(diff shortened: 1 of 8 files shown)


modules/web/js/ballerina/model/transform-factory.js
```diff
@@ -28,13 +28,19 @@ import Environment from '../env/environment';
...
      */
-    static createVariableRefExpression(name) {
+    static createVariableRefExpression(name, type) {
         const fragment = FragmentUtils.createExpressionFragment(name);
         const parsedJson = FragmentUtils.parseFragment(fragment);
         const refExpr = TreeBuilder.build(parsedJson.variable.initialExpression);
+        if (type) {
+            refExpr.symbolType = [type];
+        } else {
+            console.log('Type unknown');
```
keizer619: shoudn't we use log.error()?

maheeka: fixed

PR #4210, opened 2017-10-27
Struct generation by importing JSON

keizer619:
This feature will add a button called "Import from JSON" in the Struct Definition view.

screen shot 2017-10-27 at 10 36 26 am

User can provide a sample JSON on pop up dialog box 

screen shot 2017-10-27 at 10 37 11 am

Upon importing Struct fields will be generated from the JSON 

screen shot 2017-10-27 at 10 37 26 am

### DIFF
--- a/modules/web/js/ballerina/diagram2/views/default/components/nodes/struct-node.jsx
+++ b/modules/web/js/ballerina/diagram2/views/default/components/nodes/struct-node.jsx
@@ -407,7 +407,7 @@ class StructNode extends React.Component {
                     y={y + 10}
                     width={25}
                     height={25}
-                    className="struct-added-value-wrapper"
+                    className="struct-added-value-wrapper-light"
                 />
                 <image
                     x={x + DesignerDefaults.structDefinitionStatement.width - 30 + submitButtonPadding}
@@ -474,6 +474,7 @@ class StructNode extends React.Component {
                                         validateIdentifierName={this.validateIdentifierName}
                                         validateDefaultValue={this.validateDefaultValue}
                                         addQuotesToString={this.addQuotesToString}
+                                        index={i}
                                     />
                                 );
                             }
--- a/modules/web/js/ballerina/diagram2/views/default/components/nodes/struct-node.jsx
+++ b/modules/web/js/ballerina/diagram2/views/default/components/nodes/struct-node.jsx
@@ -214,6 +214,8 @@ class StructNode extends React.Component {
                     currentType = 'int';
                 } else if (this.isFloat(currentValue)) {
                     currentType = 'float';
+                } else if (currentValue === 'true' || currentValue === 'false') {
+                    currentType = 'boolean';
                 }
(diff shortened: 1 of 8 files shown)


modules/web/js/ballerina/diagram2/views/default/components/nodes/struct-node.jsx
```diff
@@ -187,6 +188,67 @@ class StructNode extends React.Component {
     }
 
     /**
+     * Open JSON import dialog box and pass callback
+     */
+    onClickJsonImport() {
+        const onImport = (json) => {
+            let refExpr = TreeBuilder.build(FragmentUtils.parseFragment(FragmentUtils.createExpressionFragment(json)));
+            let currentValue;
+            this.props.model.setFields([], true);
```
maheeka: Should we check for errors in building the tree. If there is an error, we shouldn't make fields empty.

modules/web/js/ballerina/diagram2/views/default/components/nodes/struct-node.jsx
```diff
@@ -187,6 +188,67 @@ class StructNode extends React.Component {
...
+            refExpr.variable.initialExpression.keyValuePairs.forEach((ketValPair) => {
```
maheeka: Also do a isRefExpr check if we are not validating the json content to be json in the text area.

modules/web/js/ballerina/diagram2/views/default/components/nodes/struct-node.jsx
```diff
@@ -187,6 +188,67 @@ class StructNode extends React.Component {
...
+            },
+        });
+    }
+
+    /**
+     * Check given value is an Integer
+     * @param  {Number} val value to check
+     * @return {Boolean}  is Intger
+     */
+    isInt(val) {
```
maheeka: Can we avoid this individual type check? Or should it be for all primitive types? Not sure what we are doing here.

modules/web/js/ballerina/diagram2/views/default/components/nodes/struct-node.jsx
keizer619: JSON will be validated from the ACE editor. Import button will be only enabled if its a valid JSON

modules/web/js/ballerina/diagram2/views/default/components/nodes/struct-node.jsx
keizer619: Sample JSON may contain 23 or 23.5  to figure out if its a Integer or Float to generate struct types these checks are used.

modules/web/js/ballerina/diagram2/views/default/components/nodes/struct-node.jsx
keizer619: As JSON is validated from ACE editor mode, there should not be errors when parsing

PR #4303, opened 2017-10-30
Visualizing public/private keyword for constructs in the design view

NatashaWso2:
Since the unlock icon is not designed yet, to depict the 'public' keyword I have used the public icon which is available in  wso2 fonts. Will include the 'unlock' icon once its designed

This fixes https://github.com/ballerinalang/composer/issues/4281


modules/web/js/ballerina/diagram2/views/default/components/decorators/panel-decorator.css
```diff
@@ -52,4 +52,10 @@
 
 .protocol-rect-http{
     fill: #3498db !important;
+}
+.publicPrivateRectHolder{
```
hemikak: Better to have class names like public-private-rect-holder since we have followed that through out the project.

modules/web/js/ballerina/diagram2/views/default/components/decorators/panel-decorator.jsx
```diff
@@ -80,6 +81,18 @@ class PanelDecorator extends React.Component {
         this.props.model.remove();
     }
 
+    togglePublicPrivateFlag() {
+        this.props.model.public = !this.props.model.public;
+        this.props.model.trigger('tree-modified', {
```
hemikak: If its to update the view, we can simply call `this.context.editor.update();` and have context types:
```
PanelDecorator.contextTypes = {
    editor: PropTypes.instanceOf(Object).isRequired,
};
```

modules/web/js/ballerina/diagram2/views/default/components/decorators/panel-decorator.jsx
```diff
@@ -156,6 +169,25 @@ class PanelDecorator extends React.Component {
...
+            TreeUtils.isStruct(this.props.model) || TreeUtils.isConnector(this.props.model)) {
+            // Toggle button for public/private flag
+            const publicPrivateFlagButtonProps = {
+                bBox: {
+                    x: x - (width * 3),
+                    y,
+                    height,
+                    width,
+                },
+                icon: ImageUtil.getSVGIconString(this.props.model.public ? 'public' : 'lock'),
```
hemikak: Can `this.props.model.public` be undefined at anytime ?
If so, `this.props.model.public === false`  will also give 'lock' and `this.props.model.public === undefined` will also give 'lock' as output.

modules/web/js/ballerina/diagram2/views/default/components/decorators/panel-decorator.jsx
NatashaWso2: We have to update the view and should do a source-gen as well. That is the reason a tree modification is done so it updates the view as well as do a source-gen.

modules/web/js/ballerina/diagram2/views/default/components/decorators/panel-decorator.jsx
NatashaWso2: this.props.model.public should be either true or false. It can never be undefined

hemikak: 👍

modules/web/js/ballerina/diagram2/views/default/components/decorators/panel-decorator.jsx
hemikak: 👍

PR #4323, opened 2017-10-31
fix anonymous struct code generation

keizer619:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/rest/datamodel/BLangFileRestService.java
```diff
@@ -190,6 +193,16 @@ public static JsonElement generateJSON(Node node) throws InvocationTargetExcepti
                 continue;
             }
 
+            if (node.getKind() == NodeKind.USER_DEFINED_TYPE && jsonName.equals("typeName")) {
+                IdentifierNode typeNode = (IdentifierNode) prop;
+                Node structNode;
+                if (typeNode.getValue().startsWith("$anonStruct$") &&
+                    (structNode = anonStructs.remove(typeNode.getValue())) != null) {
+                    nodeJson.add("anonStruct", generateJSON(structNode, anonStructs));
```
manuranga: Here you can assinge `generateJSON(structNode, anonStructs)` to a var and then `addProp("annonStruct", true)`. This way you can avoid `parent` in tree.g

PR #4338, opened 2017-11-01
 Handle runtime errors upon parse & derive program dir (if avail) during parse

kaviththiranga:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/rest/datamodel/BLangFileRestService.java
```diff
@@ -345,6 +348,25 @@ private JsonObject validateAndParse(BFile bFileRequest) throws InvocationTargetE
             result.add("packageInfo", packageInfo);
         }
 
+        if(bFileRequest.needProgramDir() && (model != null && model.pkgDecl != null)) {
```
maheeka: This will cause a  checkstyle issue

PR #4352, opened 2017-11-02
Modified UI for try it. Fixes for try-it.

hemikak:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/rest/TryItService.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2016, WSO2 Inc. (http://wso2.com) All Rights Reserved.
+ * Copyright (c) 2016-2017, WSO2 Inc. (http://wso2.com) All Rights Reserved.
```
srsharon: This should be just 2016

modules/web/src/plugins/try-it/views/http-client.scss
```diff
@@ -225,63 +219,185 @@ hr{
...
+            .react-autosuggest__container {
+                .inputContainer {
+                    height: 33px;
+                    padding: 6px;
+                    border: none;
+                    border-bottom: 1px solid #3b3b3b;
+                    input[type=text]{
+                        color: #fff;
+                        background-color: #333;
+                        border: 0px solid #333;
```
srsharon: remove px from 0

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/rest/TryItService.java
hemikak: Fixed in https://github.com/ballerinalang/composer/pull/4354

modules/web/src/plugins/try-it/views/http-client.scss
hemikak: Fixed in https://github.com/ballerinalang/composer/pull/4354

PR #4377, opened 2017-11-03
Fix the error with globals cannot be added

NipunaMarcus:


modules/web/js/ballerina/diagram2/views/default/components/nodes/top-level-nodes.jsx
```diff
@@ -164,7 +164,7 @@ class TopLevelNodes extends React.Component {
         if (!value) {
             return;
         }
-        value = `${value};\n`;
+        value = `${value}\n`;
```
hevayo: I think better to check if the semicolon is there and append based on that.

NipunaMarcus: semicolon will always be there @hevayo. It is getting attached from expression editor.

PR #4389, opened 2017-11-03
 Fixing positioning issue with globals and imports

NatashaWso2:


modules/web/js/ballerina/diagram2/views/default/positioning-util.js
```diff
@@ -260,6 +260,25 @@ class PositioningUtil {
                 node.viewState.components.globalsBbox.x += packageDefTextWidth;
             }
         }
+
+        if (node.viewState.importsExpanded) {
+            const importDeclarationHeight = 30;
+            const importInputHeight = 40;
+            const topBarHeight = 25;
+            const yGutterSize = 10;
```
nadeeshaan: is it possible to move these constants to the designer defaults?

NatashaWso2: Fixed with daa1724

PR #4394, opened 2017-11-03
fix transformer dropdown type loading and source target selection

keizer619:


modules/web/js/ballerina/diagram2/views/default/components/transform/transform-expanded.jsx
```diff
@@ -922,8 +921,9 @@ class TransformExpanded extends React.Component {
      * @memberof TransformExpanded
      */
     addSource(source) {
-        if (this.isVertexExist(source)) {
-            this.props.model.addInput(source);
+        const vertex = this.state.vertices.filter((val) => { return val === source; })[0];
```
maheeka: are we using vertices?

keizer619: @maheeka 
types are loaded as vertices into drop downs and variable will be created upon selection and add to source/target

PR #4402, opened 2017-11-03
Make transformer name edittable

roonyh:
final

### DIFF
--- a/modules/web/js/ballerina/diagram2/views/default/sizing-util.js
+++ b/modules/web/js/ballerina/diagram2/views/default/sizing-util.js
@@ -837,6 +837,32 @@ class SizingUtil {
 
         const textWidth = this.getTextWidth(node.getSignature());
         viewState.titleWidth = textWidth.w;
+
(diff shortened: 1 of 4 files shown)


modules/web/js/ballerina/diagram2/views/default/sizing-util.js
```diff
@@ -837,6 +837,32 @@ class SizingUtil {
...
         viewState.titleWidth = textWidth.w;
+
+        const returnParams = _.join(node.getReturnParameters().map(ret => ret.getSource()), ',');
+        const typeText = `< ${node.getSourceParam().getSource()}, ${returnParams} >`;
+        const typeTextDetails = this.getTextWidth(typeText, 0);
+        viewState.typeText = typeTextDetails.text;
+        viewState.typeTextWidth = typeTextDetails.w;
+
+        const params = _.join(node.getParameters().map(ret => ret.getSource()), ',');
+        const paramText = `( ${params} )`;
```
maheeka: if there is no name () must not be shown. Refer getSignature() in transformer-node

PR #4439, opened 2017-11-04
fix add variable and parameter support to new transformer

keizer619:


modules/web/js/ballerina/diagram2/views/default/components/transform/transform-expanded.jsx
```diff
@@ -1153,27 +1170,24 @@ class TransformExpanded extends React.Component {
                 return this.getVerticeData(varN);
             });
         } else {
+            const variableType = {};
+            if (varNode.getKind() === 'VariableDef') {
```
roonyh: Better use TreeUtil.isVariableDef. But not a blocker 🙂

Issue #4404, opened 2017-11-04
No tooltip for copy URL in try-it


dilinisg:
![2](https://user-images.githubusercontent.com/1845370/32401770-df889b3e-c13b-11e7-919f-64e7777de0f4.png)

PR #4530, opened 2017-11-07
Added color coding for response attributes depending on status code. Allowed custom http methods.

hemikak:


hemikak:
Fixes https://github.com/ballerinalang/composer/issues/4404

PR #4535, opened 2017-11-07
Improve left panel: Add toggle shortcut

kaviththiranga:


modules/web/src/core/layout/components/LeftPanel.jsx
```diff
@@ -180,7 +180,9 @@ class LeftPanel extends React.Component {
         const tabs = [];
         const panes = [];
         const { views, onActiveViewChange, ...restProps } = this.props;
-        const activeViewPrev = this.props.activeView || this.context.history.get(HISTORY.ACTIVE_LEFT_PANEL_VIEW);
+        const activeViewPrev = this.props.activeView
+                    || this.context.history.get(HISTORY.ACTIVE_LEFT_PANEL_VIEW)
+                    || views[0].id;
```
hemikak: Will there always be atleast one view ?

kaviththiranga: yeah in our case, explorer, run will be there. But better to fix for all. I have added the fix. Thanks,

Issue #3514, opened 2017-10-11
Semicolon is not ignored when adding the package name


dilinisg:
1. Add a package name as `org.test;`

**Issue**
Gives an error because two semicolons are added. Semicolon is not ignored when adding the package name like in other variables/assignments

PR #4541, opened 2017-11-08
Fixing issues when naming structs and adding package name with semi-colon

NatashaWso2:
Fixes https://github.com/ballerinalang/composer/issues/3514

### DIFF
--- a/modules/web/js/ballerina/diagram2/views/default/components/nodes/top-level-nodes.jsx
+++ b/modules/web/js/ballerina/diagram2/views/default/components/nodes/top-level-nodes.jsx
@@ -96,9 +96,7 @@ class TopLevelNodes extends React.Component {
         }
         if (this.packageDefValue) {
             // If there is a semi-colon at the end of the package name, remove it
-            if (this.packageDefValue.slice(-1) === ';') {
-                this.packageDefValue = this.packageDefValue.slice(0, -1);
-            }
+            this.packageDefValue = _.trimEnd(this.packageDefValue, ';');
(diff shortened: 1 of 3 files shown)


modules/web/js/ballerina/diagram2/views/default/components/nodes/top-level-nodes.jsx
```diff
@@ -95,6 +95,10 @@ class TopLevelNodes extends React.Component {
             this.props.model.viewState.packageDefExpanded = false;
         }
         if (this.packageDefValue) {
+            // If there is a semi-colon at the end of the package name, remove it
+            if (this.packageDefValue.slice(-1) === ';') {
```
hemikak: You can use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/endsWith function here.

hemikak: Or use this.packageDefValue = _.trimEnd(this.packageDefValue, ';') , so  you can get rid of this whole block.

NatashaWso2: Used the _trimEnd function of Lodash  https://lodash.com/docs/4.17.4#trimEnd 
Fixed with commit 05f4d18

NatashaWso2: Fixed with commit 05f4d18

PR #4542, opened 2017-11-08
Go to file in explorer upon save as

kaviththiranga:


modules/web/src/core/workspace/plugin.js
```diff
@@ -151,6 +162,20 @@ class WorkspacePlugin extends Plugin {
...
 
     /**
+     * Checks whether the given path is opened in explorer
+     * @param {String} filePath target path
+     */
+    isFilePathOpenedInExplorer(filePath) {
+        let folderOpened = false;
+        this.openedFolders.forEach((folder) => {
+            if (filePath.startsWith(folder)) {
+                folderOpened = true;
```
hemikak: Should there be a break here ?

hemikak: ignore this.

PR #4582, opened 2017-11-13
Add issue template into composer repo

dakshika:
Add issue template into composer repo


hemikak:
Should we have a separate place to put any logs found ?

Issue #3652, opened 2017-10-12
Stack-overflow error when the swagger is edited


yasassri:
*Steps*

1. Create a service.
2. Go to swagger view.
3. Manually change an element in the swagger.
4. Click on design view

**Issue:** Following StackOverflow error is thrown.

```
Oct 12, 2017 7:19:01 PM org.wso2.msf4j.internal.MSF4JMessageProcessor handleThrowable
WARNING: Unmapped exception
java.lang.StackOverflowError
	at org.wso2.ballerinalang.compiler.parser.BLangPackageBuilder.addImportPackageDeclaration(BLangPackageBuilder.java:859)
	at org.wso2.ballerinalang.compiler.parser.BLangParserListener.exitImportDeclaration(BLangParserListener.java:140)
	at org.wso2.ballerinalang.compiler.parser.antlr4.BallerinaParser$ImportDeclarationContext.exitRule(BallerinaParser.java:519)
	at org.antlr.v4.runtime.Parser.triggerExitRuleEvent(Parser.java:430)
	at org.antlr.v4.runtime.Parser.exitRule(Parser.java:646)
	at org.wso2.ballerinalang.compiler.parser.antlr4.BallerinaParser.compilationUnit(BallerinaParser.java:336)
	at org.wso2.ballerinalang.compiler.parser.Parser.generateCompilationUnit(Parser.java:96)
	at org.wso2.ballerinalang.compiler.parser.Parser.lambda$parse$7(Parser.java:75)
	at java.util.ArrayList.forEach(ArrayList.java:1249)
	at org.wso2.ballerinalang.compiler.parser.Parser.parse(Parser.java:75)
	at org.wso2.ballerinalang.compiler.PackageLoader.sourceCompile(PackageLoader.java:161)
	at org.wso2.ballerinalang.compiler.PackageLoader.loadPackage(PackageLoader.java:147)
	at org.wso2.ballerinalang.compiler.PackageLoader.loadPackageNode(PackageLoader.java:125)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolEnter.visit(SymbolEnter.java:252)
	at org.wso2.ballerinalang.compiler.tree.BLangImportPackage.accept(BLangImportPackage.java:72)
	at org.wso2.ballerinalang.compiler.semantics.analyzer.SymbolEnter.defineNode(SymbolEnter.java:157)
[...]

PR #4595, opened 2017-11-14
Bug fixes

maheeka:
This PR contains,

Generic fix for https://github.com/ballerinalang/composer/issues/3652 which cannot be reproduced


modules/web/src/plugins/export-diagram/commands.js
```diff
@@ -26,6 +26,9 @@ export function getCommandDefinitions(plugin) {
     return [
         {
             id: COMMANDS.SHOW_EXPORT_DIAGRAM_DIALOG,
+            shortcut: {
+                default: 'shift+d',
```
kaviththiranga: doesn't it type D

PR #4599, opened 2017-11-14
Add Connector action completion tests

nadeeshaan:


hemikak:
Any reason to keep the expected value files as "*.exp" and not "*.json" .

Having json would be good IMO, as it will be detected by IDEs as well.

nadeeshaan:
We are maintaining two files with the same name and different extensions (.bal and .exp). Added the .exp extension in order to the file being descriptive. Also, this is the initial framework implementation and will change this format in the next iteration of the framework implementation.

hemikak:
👍

Issue #3984, opened 2017-10-19
Run Icon in the left pane is confusing


yasassri:
Can we change this icon and the tooltip since this doesn't really run the program?

 
![screenshot from 2017-10-19 16-23-49](https://user-images.githubusercontent.com/7681361/31767771-10414ed2-b4ea-11e7-83db-52410447f2a0.png)

maheeka:
Changed the wordings of the debugger panel as below. 
![image](https://user-images.githubusercontent.com/1029806/32825724-ebd6f2dc-c9dd-11e7-8aab-f0b7e07b79a5.png)

PR #4600, opened 2017-11-14
Change debugger panel run wordings to debug

maheeka:
Before : 
image
Now : 
image
Resolves : https://github.com/ballerinalang/composer/issues/3984


modules/web/src/plugins/debugger/plugin.js
```diff
@@ -110,16 +110,16 @@ class DebuggerPlugin extends Plugin {
...
                     // region specific options for left-panel views
                     regionOptions: {
-                        activityBarIcon: 'start',
-                        panelTitle: 'Run',
+                        activityBarIcon: 'bug',
+                        panelTitle: 'Debugger',
                         panelActions: [
                             {
-                                icon: 'configarations',
+                                icon: 'configurations',
```
pahans: This is intensionally misspelled because it is a typo in font wso2. It is corrected in the font wso2 repo now. We should correct this when we get font Wso2 latest

roonyh: Since we build our font now we should just rename this file https://github.com/ballerinalang/composer/blob/master/modules/web/font/font-ballerina/icons/configarations.svg

@dakshika could you please confirm?

maheeka: I have renamed the svg file as well with this commit. So this should work

Issue #2245, opened 2017-07-02
Navigate to Composer browser window upon the composer start


sumuditha-viraj:
Open/navigate composer in browser automatically once we start the composer from command line.

dakshika:
Are we supporting for edge browser? By default, Edge is the default browser for Windows. so If we are implementing this ,we need to check against the all respective default browsers.

PR #4607, opened 2017-11-15
Navigate to composer browser window when the composer starts

NatashaWso2:
Works with : Linux, Windows, Mac
Fixes https://github.com/ballerinalang/composer/issues/2245

### DIFF
--- a/modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/app/StartBrowser.java
+++ b/modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/app/StartBrowser.java
@@ -1,8 +1,19 @@
-package org.ballerinalang.composer.service.workspace.app;
-
-/**
- * Created by natasha on 11/15/17.
+/*
+ * Copyright (c) 2016, WSO2 Inc. (http://wso2.com) All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/app/StartBrowser.java
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Starting the default browser once the composer starts
+ */
+public class StartBrowser {
+
+    public static void startInDefaultBrowser(String url) throws IOException {
+        Locale locale = Locale.ENGLISH;
+        String os = System.getProperty("os.name").toLowerCase(locale);
+        Runtime rt = Runtime.getRuntime();
+            if (os.indexOf("win") >= 0) {
```
pahans: you can use LaunchUtils.isWindows method to check this https://github.com/ballerinalang/composer/blob/master/modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/launcher/util/LaunchUtils.java#L32

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/app/StartBrowser.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (c) 2016, WSO2 Inc. (http://wso2.com) All Rights Reserved.
```
hemikak: Year is wrong.

NatashaWso2: Fixed with fbbb851

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/app/StartBrowser.java
NatashaWso2: Fixed with fbbb851

PR #4630, opened 2017-11-16
Generate default constructs to a js file

pahans:
Improves drag and drop performance


modules/web/gen-default-nodes/index.js
```diff
@@ -0,0 +1,71 @@
...
+import fs from 'fs';
+import defToFragment from './def-to-fragment';
+
+/**
+ * parse fragment asynchronously
+ * @param {string} fragment - source fragment.
+ * @return {object} fragment details to be sent to fragment parser.
+ */
+
+const ENDPOINT = 'http://localhost:8289/ballerina/model/parse-fragment';
```
maheeka: Can we use https://github.com/ballerinalang/composer/blob/e70f81097763907c07aed2d7efc19fdb32aaa06d/modules/web/js/api-client/api-client.js#L34 to get the endpoint?

modules/web/gen-default-nodes/def-to-fragment.js
```diff
@@ -0,0 +1,299 @@
...
+        return NodeTypes.createTopLevelNodeFragment(`
+            annotation annotation1 {
+            }
+        `);
+    },
+    createAssignmentStmt: () => {
+        return NodeTypes.createStatementFragment('var a = 1;');
+    },
+    createBindStmt: () => {
+        return NodeTypes.createStatementFragment('bind __connector with __endpoint;');
```
maheeka: Not related to your fix, but shall we also change this to `bind _connector with _endpoint;` ? Earlier we had a concern raised for using double underscore as variable prefix.

modules/web/gen-default-nodes/index.js
pahans: we cannot import `api-client` here. Because we run this file in a command line environment. `api-client` has some webpack alias's (event_channel).

modules/web/js/ballerina/utils/fragment-utils/index.js
```diff
@@ -0,0 +1,40 @@
+/*
```
hevayo: We currently don't use index.js in our code so shall we stick with a proper file name.

modules/web/gen-default-nodes/index.js
```diff
@@ -0,0 +1,86 @@
+/**
```
hevayo: We don't use the index.js in other places so shall we stick with a proper file name

modules/web/js/ballerina/utils/fragment-utils/index.js
pahans: I have separated `fragment-utils.js` file into a submodule. index.js is intended to expose the API. importing fragments-utils is same as before.

PR #4633, opened 2017-11-17
Set package alias when dropping function invocations

NatashaWso2:


modules/web/js/ballerina/model/default-node-factory.js
```diff
@@ -568,6 +568,9 @@ class DefaultNodeFactory {
             node.setVariables(returnNode.getVariables());
         }
         node.getExpression().getName().setValue(functionDef.getName());
+         if (packageName && packageName !== 'Current Package' && packageName !== "builtin" ) {
```
maheeka: use single quote for 'builtin' to avoid eslint errors

modules/web/js/ballerina/model/default-node-factory.js
```diff
@@ -551,7 +551,7 @@ class DefaultNodeFactory {
         if (functionDef.getParameters()) {
            let parameters = [];
            functionDef.getParameters().map((param) => {
-               let defaultValue = Environment.getDefaultValue(param.type) || null;
+               let defaultValue = Environment.getDefaultValue(param.type) || 'null';
```
nadeeshaan: any specific reason for having null as a string?

modules/web/js/ballerina/model/default-node-factory.js
nadeeshaan: use single quotes instead of double quotes

modules/web/js/ballerina/model/default-node-factory.js
maheeka: This is to add null as the parameter. We have to set it here. Maybe we can change the getDefaultValue to check and return null in order to avoid redundant logic. @NatashaWso2 shall we do that?

NatashaWso2: The getDefaultValue() checks and returns null if the default value is undefined or nil

modules/web/js/ballerina/model/default-node-factory.js
NatashaWso2: Fixed with 6dbe0dd89c23f52172acc6bc2be1a8a3d69fa440

NatashaWso2: Fixed with 6dbe0dd89c23f52172acc6bc2be1a8a3d69fa440

PR #4649, opened 2017-11-17
Modified logging for the composer.

hemikak:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/rest/WorkspaceService.java
```diff
@@ -82,7 +82,7 @@ public Response root(@DefaultValue(".bal") @QueryParam("extensions") String exte
                     .type(MediaType.APPLICATION_JSON)
                     .build();
         } catch (Throwable throwable) {
-            logger.error("/root service error", throwable.getMessage());
+            logger.error("/root service error", throwable.getMessage(), throwable);
```
maheeka: @hemikak wouldn't this print the java stacktrace in console?

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/rest/WorkspaceService.java
```diff
@@ -266,7 +266,7 @@ public Response userHome() {
                     .entity(workspace.getUserHome()).header("Access-Control-Allow-Origin", '*')
                     .type(MediaType.APPLICATION_JSON).build();
         } catch (Throwable throwable) {
-            logger.error("/userHome service error", throwable.getMessage());
+            logger.error("/userHome service error", throwable.getMessage(), throwable);
```
maheeka: @hemikak doesn't this print Java stacktrace in console?

PR #4652, opened 2017-11-17
Fix cannot debug package issue, fixes #3893

### LINKED ISSUE #3893: Debug doesn't work when packaged
**Steps**

1. Create following 2 bal files with a package called package2.

bal1.bal

```
package package2;
import ballerina.lang.system;
function main (string[] args) {
    system:println("1111");
    func();
    system:println("2222");
}
```

bal2.bal

```
package package2;
import ballerina.lang.system;
function func () {
    system:println("INNNNNN222NN");
}
```
2. Add few debug points to each file.
3 Run debug.

**Issue:** Debug points do not get hit.

pahans:


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/launcher/Command.java
```diff
@@ -157,6 +142,22 @@ public void setCommandArgs(String[] commandArgs) {
                 }
             }
         }
+
+        if (packagePath == null) {
```
NipunaMarcus: can we use === instead of ==?

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/launcher/Command.java
```diff
@@ -157,6 +142,22 @@ public void setCommandArgs(String[] commandArgs) {
...
+        } else {
+            commandList.add(packagePath);
+        }
+
+        if (debug) {
+            commandList.add("--ballerina.debug");
+            commandList.add(String.valueOf(this.port));
+        }
+
+        if (this.commandArgs != null) {
```
NipunaMarcus: can we use !== instead of !=?

Issue #4583, opened 2017-11-13
parameter for file location 


robbla:
Would it be possible to have a parameter for the file location that composer opens on by default? 

Either as a startup parameter, part of the sh or bat file or even just in a settings or config file

setting this will allow a different view, pointing to a directory where .bal files are stored and make work faster

pahans:
From next composer release onwards, you can pass --file param to composer script to open a file by default. We will work on starting up with a default directory too.

PR #4667, opened 2017-11-19
Open startup file if param passed to composer, fixes #4523

### LINKED ISSUE #4523: composer should load sample service when start up
Right now the user gets in the first run is the welcome page.

pahans:


NipunaMarcus:
This is also related to #4583

Issue #3961, opened 2017-10-19
Connectors are not adjusted according to the length of the connector name


dilinisg:
Connectors are not adjusted according to the length of the connector name 
![7](https://user-images.githubusercontent.com/1845370/31758277-69d1334c-b4ca-11e7-883e-a0ca59e0ada3.png)

PR #4693, opened 2017-11-22
Truncating the connector name according to the endpoint construct width

NatashaWso2:
Fixes https://github.com/ballerinalang/composer/issues/3961


modules/web/js/ballerina/diagram2/views/default/sizing-util.js
```diff
@@ -1587,6 +1587,11 @@ class SizingUtil {
         const viewState = node.viewState;
         this.sizeStatement(node.getSource(), viewState);
         this.adjustToLambdaSize(node, viewState);
+
+        // Truncate the endpoint name to fit the statement box
+        if (TreeUtil.isEndpointTypeVariableDef(node)) {
+            viewState.endpointIdentifier = this.getTextWidth(node.getVariable().getName().value, 0, 100).text;
```
nadeeshaan: Shall we get the value 100 to a meaningful variable and use it? Otherwise, it would be bit difficult to understand what the value meant to be.

NatashaWso2: Fixed with 52c36d792ace3d77c1130ff45c3456ea660babc4

Issue #4705, opened 2017-11-23
Enum support for Connector properties view.


hevayo:
SQL connector will be changed to accept enums for some of the configuration properties. see https://github.com/ballerinalang/ballerina/issues/4100

We need to modify the properties window of the connector to support enums in composer.

PR #4719, opened 2017-11-23
 Enum support for Connector properties view

NatashaWso2:
Fixes https://github.com/ballerinalang/composer/issues/4705


modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/model/Enum.java
```diff
@@ -0,0 +1,54 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ballerinalang.composer.service.workspace.langserver.model;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Enum.
```
hemikak: A better class comment would be better

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/model/Enumerator.java
```diff
@@ -0,0 +1,54 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.ballerinalang.composer.service.workspace.langserver.model;
+
+/**
+ * Enumerator.
```
hemikak: A better class comment would be good.

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/model/Enum.java
NatashaWso2: Fixed with 0aa1ac6

modules/services/workspace-service/src/main/java/org/ballerinalang/composer/service/workspace/langserver/model/Enumerator.java
NatashaWso2: Fixed with 0aa1ac6

modules/web/js/ballerina/env/enum.js
```diff
@@ -0,0 +1,144 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+import _ from 'lodash';
+import BallerinaEnvFactory from './ballerina-env-factory';
+
+/**
+ * @class Enum
+ * @augments
+ * @param {Object} args data to create the enum
```
nadeeshaan: args are being duplicated in the constructor too

modules/web/js/ballerina/diagram/views/default/components/utils/connector-properties.jsx
```diff
@@ -256,7 +258,18 @@ class ConnectorPropertiesForm extends React.Component {
             if (key.bType === 'string') {
                 key.value = this.addQuotationForStringValues(key.value);
             }
-            paramArray.push(key.value);
+            if (key.bType === 'enum') {
+                let enumValue = 'null';
```
NipunaMarcus: why are we putting null as a string? is it coming from the backend?

modules/web/js/ballerina/env/enum.js
NatashaWso2: Fixed with 533f6ba5b5a478b16da7f64f7acdafaf47712d8d

modules/web/js/ballerina/diagram/views/default/components/utils/connector-properties.jsx
NatashaWso2: No its not coming from the backend. Since there's no default value defined for the enum atm, the default value for the enum would be null i.e. if the user does not select any value, null would be taken as the value.

PR #4744, opened 2017-11-28
Improve workspace folder handling with models

kaviththiranga:
Previously we only handled opened folders via
the full path and we didn't have any model object
representing the opened folder in workspace. For
program dir related improvements for composer, we
need some attributes for folders opened in workspace.

### DIFF
--- a/modules/web/src/core/workspace/components/ExplorerItem.jsx
+++ b/modules/web/src/core/workspace/components/ExplorerItem.jsx
@@ -19,16 +19,13 @@
 
 import React from 'react';
 import classnames from 'classnames';
-import _ from 'lodash';
-import { getPathSeperator } from 'api-client/api-client';
 import PropTypes from 'prop-types';
-import { COMMANDS } from './../constants';
-import Folder from './../model/folder'
 import ContextMenuTrigger from 'core/view/context-menu/ContextMenuTrigger';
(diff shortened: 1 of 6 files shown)


modules/web/src/core/workspace/components/ExplorerItem.jsx
```diff
@@ -23,11 +23,12 @@ import _ from 'lodash';
 import { getPathSeperator } from 'api-client/api-client';
 import PropTypes from 'prop-types';
 import { COMMANDS } from './../constants';
-import ContextMenuTrigger from './../../view/context-menu/ContextMenuTrigger';
+import Folder from './../model/folder'
```
pahans: change this to ../

modules/web/src/core/workspace/model/folder.js
```diff
@@ -0,0 +1,120 @@
...
+
+import _ from 'lodash';
+import { getPathSeperator } from 'api-client/api-client';
+import EventChannel from 'event_channel';
+import { EVENTS } from './../constants';
+
+/**
+ * Represents a Folder opened in workspace
+ */
+class Folder extends EventChannel {
```
hevayo: Do we need an event channel for the folder, will it make more sense to handle all the event via app event stream.

PR #4753, opened 2017-11-29
Add Travis config file

Shan1024:


.travis.yml
```diff
@@ -0,0 +1,12 @@
+language: java
+
+install:
+  - . $HOME/.nvm/nvm.sh
+  - nvm install 8.9.0
+  - nvm use 8.9.0
+  - mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```
pahans: why are we skipping tests?

Shan1024: Tests are executed in a later phase using `mvn test`.

PR #4798, opened 2017-12-06
Adding readme on - how to do source gen

manuranga:


manuranga:
duplicates https://github.com/ballerinalang/composer/pull/4800

Issue #4908, opened 2018-01-02
Go-to-source does not scroll when line is not visible (monaco revamp)


kaviththiranga:

kaviththiranga:
Fixed with https://github.com/ballerinalang/composer/pull/4909

PR #4909, opened 2018-01-02
Fix duplicate folder open popup for pacakged samples and make go-to-source reveal hidden lines

kaviththiranga:
Fixes https://github.com/ballerinalang/composer/issues/4908


hemikak:
@kaviththiranga , The purpose of 'isFile' was to state that it is a packaged sample(not a single file) and the samples on the welcome would have a different view(stacked). Wouldnt this change the view of the samples in the welcome page. ?

kaviththiranga:
@hemikak 

Seems that same property has been used to file a folder open for the parent folder. shall we introduce another property then? I didn't notice the changes to view.

hemikak:
@kaviththiranga , yes, lets change the properties. +1

kaviththiranga:
@hemikak updated, plz review.

PR #4920, opened 2018-01-03
Migrate to react 16

pahans:


modules/web/src/plugins/ballerina/utils/react-try-catch-batching-strategy.js
```diff
@@ -1,50 +0,0 @@
-/**
```
hevayo: Shall we add error boundaries to ballerina file editor and diagram.

pahans: @hevayo pushed Error Boundaries component
selection_096

PR #4921, opened 2018-01-05
improve menu button sizing and positioning

keizer619:


modules/web/src/plugins/ballerina/interactions/button.jsx
```diff
@@ -57,7 +57,15 @@ class Button extends React.Component {
                             <i style={{ color: btnIconColor }} className={`fw fw-${this.props.icon} fw-stack-1x`} />
                         </span>
                     </div>
-                    {this.props.children}
+                    <div style={{
+                        left: btnRadius + 2 + (btnRadius / 4),
```
nadeeshaan: what if we precalculate these values instead of defining inline

PR #4952, opened 2018-01-12
adding if icon to the interaction menu

srsharon:


modules/web/font/font-ballerina/icons/if-else.svg
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="utf-8"?>
+
```
dakshika: remove unwanted spaces

PR #4956, opened 2018-01-17
Revamp try-catch-finally representation

maheeka:
This PR includes the new try-catch-finally design implementation.  Refer screenshots below : 
image
image
Nested scenarios : 
image


modules/web/src/plugins/ballerina/diagram/views/default/components/nodes/try-statement-decorator.jsx
```diff
@@ -0,0 +1,444 @@
...
+                <ActionBox
+                    bBox={actionBoxBbox}
+                    show={this.state.active}
+                    isBreakpoint={isBreakpoint}
+                    onDelete={() => this.onDelete()}
+                    onJumptoCodeLine={() => this.onJumpToCodeLine()}
+                    onBreakpointClick={() => this.props.onBreakpointClick()}
+                    disableButtons={this.props.disableButtons}
+                />
+                {(() => {
```
pahans: This render function is too long and difficult to read? shall we break this into several functions?

maheeka: This includes some inter-dependent code on drawing components related to catch in try catch. i.e. drawing connector lines to finally block and expanding the vertical lifeline from catch up to the 'finally' block which requires manipulating the coordinates. Therefore, it is not feasible to move this logic to a different function.

PR #4959, opened 2018-01-17
Go to definition

pahans:


modules/web/src/plugins/ballerina/views/source-editor.jsx
```diff
@@ -193,6 +193,34 @@ class SourceEditor extends React.Component {
                         },
                     },
                 });
+
+                editorInstance.onMouseDown((e) => {
```
hevayo: Does this send duplicate messages on click ? assuming Monaco also send a request for mouse down.

pahans: `e.event.preventDefault();` prevents from doing so.

PR #4963, opened 2018-01-19
Add buttons on block nodes

hevayo:
Add item add button to all the block nodes. 
screenshot from 2018-01-19 08-10-18


modules/web/src/plugins/ballerina/diagram/views/default/components/controllers/compilation-unit-ctrl.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+    /**
+     * Render Function for the Next statement.
+     * @return {React.Component} next node react component.
+     * */
+    render() {
+        const node = this.props.model;
+        const offsetX = node.getPackageDeclaration()
+                || node.viewState.packageDefExpanded
+                ? 320
```
maheeka: shouldn't we get these from designer defaults?

PR #4996, opened 2018-01-25
Fix backend url for default node generator

pahans:


modules/web/tools/gen-default-nodes/gen-default-nodes.js
```diff
@@ -58,7 +58,8 @@ function getConfigs() {
  */
 function generateNodes() {
     getConfigs().then((configs) => {
-        const { endpoint } = configs.services.fragmentParser;
+        const parserEndpoint = configs.services['ballerina-parser'].endpoint;
```
NipunaMarcus: do we need the "parserEndpoint" variable?

pahans: It's not mandatory. But I think it makes it easier to understand the code.

PR #5000, opened 2018-01-25
Make sure back end is up before attempting js tests.

roonyh:
Previously the back end composer services was started as a async maven
execution. The services may or may not be up when starting js tests.

By starting them inside the test script itself and waiting for services
stdout to print the success message, its guaranteed that the back end is
up when starting js tests.

### DIFF
--- a/modules/server/server-launcher/src/main/java/org/ballerinalang/composer/server/launcher/ServerLauncher.java
+++ b/modules/server/server-launcher/src/main/java/org/ballerinalang/composer/server/launcher/ServerLauncher.java
@@ -138,7 +138,6 @@ public class ServerLauncher {
         try {
             server.start();
             String uRL = "http://" + config.getHost() + ":" + config.getPort();
-            logger.info("Composer started successfully at " + uRL);
             PrintStream out = System.out;
             out.println("Composer started successfully at " + uRL);
             BrowserLauncher.startInDefaultBrowser(uRL);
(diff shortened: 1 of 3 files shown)


modules/server/server-launcher/src/main/java/org/ballerinalang/composer/server/launcher/ServerLauncher.java
```diff
@@ -139,6 +139,8 @@ public static void main(String[] args) {
             server.start();
             String uRL = "http://" + config.getHost() + ":" + config.getPort();
             logger.info("Composer started successfully at " + uRL);
```
hevayo: Shall we remove the log infor otherwise do the message duplicate ?

roonyh: It didn't show up in the stdout. Is it logged to a log file? Let me check again.

roonyh: Message is not duplicated in the stdout. logger puts it in a log file I think.
capture

hevayo: Shall we remove it since it is redundant. it doesn't make sense in log file to have that

roonyh: Sure. Removed in 4e2fe90

PR #5001, opened 2018-01-25
Change log levels of config reading errors

kaviththiranga:
These errors should not be visible unless user wants to see debug level logs.


kaviththiranga:
ATM if a well formatted config file is not there, everyone sees an error message. This is to avoid that

